PR #240, opened 2026-04-27
docs(troubleshooting): emulator grey-screen recovery via qcow2 reset

BenGWeeks:
## Summary

Documents a recurring emulator failure mode that the existing TROUBLESHOOTING entries didn't cover. Found while trying to capture screenshots for #234 / #238 — burned ~2 hours on it before identifying the actual cause.

## What's the failure mode

- Emulator boots far enough that `adb devices` sees it, but Android userspace never reaches `getprop sys.boot_completed=1`
- `adb` reports the device as `offline` indefinitely, sometimes flapping between `device` and `offline` as `system_server` watchdog fires
- Visual: grey QEMU window, GPU init succeeds in the emulator log (`Setting display: 0 configuration to: 1080x2400`, `Created extended window`) but every subsequent boot-completed poll fails with `device offline`

## Cause

`~/.android/avd/<AVD_NAME>.avd/userdata-qemu.img.qcow2` accumulates corrupt state from prior crashed boots (Digital Wellbeing ANR cascades, hard-killed `qemu-system`, Vulkan/gfxstream stalls). Once corrupt, the corruption re-mounts on every cold boot — no GPU mode switch, RAM bump, or `adb reboot` helps. The file balloons to 7 GB+ over time.

## Fix

Factory-reset just the runtime state — delete the qcow2 overlays + snapshots + lock files, leave the AVD config and base `.img` files alone. Equivalent to recreating the AVD without needing `avdmanager` / `cmdline-tools` installed.

## False leads documented

Saved future runs from re-burning time on:
- `adb reboot` (corrupt overlay re-mounts on every boot)
- Switching `-gpu host` → `-gpu swiftshader_indirect` (GPU init isn't where it fails)
- Increasing AVD RAM (`system_server` isn't OOM)
- Re-running `expo prebuild` (this is emulator state, not app build state)

## Test plan

- [x] Followed the documented recipe end-to-end on this machine; emulator went from "stuck grey screen, 3 boot attempts failed" → "clean boot in <30 s" with hardware GPU.
- [x] AsciiDoc renders correctly (no broken table).


BenGWeeks (commented):
## Self-review

### Strengths

- **Clear failure-mode signature**: the symptom (grey screen, `boot_completed` never flips, ADB flapping `device` ↔ `offline`) is distinctive enough that someone hitting it can recognise the entry by the headline alone.
- **The false-leads list is genuinely useful** — every item corresponds to a path that was actually wasted today (`adb reboot`, GPU mode swap, RAM bump, `expo prebuild`). Saves ~10 min of re-discovery for the next person.
- **The `cp config.ini config.ini.preboot-bak` step** correctly references the existing TROUBLESHOOTING entry's `sed -i`-truncates-config gotcha. Cross-references are doing their job.
- **The fix scope is correctly narrow** — it deletes only the qcow2 overlays + snapshots + lock files, leaving the AVD config and base `*.img` files untouched. Doesn't require `cmdline-tools` or `avdmanager` recreate.

### Minor

- The recipe says "the userdata-qemu.img.qcow2 file goes from a multi-GB accumulated overlay (often 7 GB+ after months of use) back to a thin overlay." Worth noting the disk-space win as a secondary benefit — recovers ~7 GB on this run.
- The `What you lose` callout is good. Could also mention that `eas build --local` from PR #241's path still works fine post-reset since EAS pulls everything from servers.

### Concerns

None blocking. The recipe is correct and the symptom-cause-fix structure follows the existing TROUBLESHOOTING.adoc convention.

### Conflict watch

No conflict with PR #241 / #242 — different file (`TROUBLESHOOTING.adoc` vs `DEPLOYMENT.adoc`).

PR #241, opened 2026-04-27
docs(deployment): local prod build recipe for sideloading over EAS install

BenGWeeks:
## Summary

Documents the workflow for sideloading a release APK onto a Pixel that already has an EAS-built production install — needed to preserve user data (wallets, Nostr login, message history) when validating a release locally.

## Problem this fixes

Two unrelated install failures stack on top of each other when you run `npx expo run:android --variant release`:

1. **Signature mismatch** — locally-signed APKs and EAS-signed APKs use different keystores; the package manager refuses cross-keystore upgrades with `INSTALL_FAILED_UPDATE_INCOMPATIBLE`. The `-r -d` flags don't help.
2. **Version downgrade** — local prebuilds default `android.versionCode` to `1`, while EAS's remote autoIncrement counter (`appVersionSource: "remote"` in `eas.json`) is in the high 20s. Even if you fix the signature, install fails with `INSTALL_FAILED_VERSION_DOWNGRADE`.

Hit both errors today while trying to release the merged #234 + #238 to my Pixel — without docs, took ~20 min of re-discovery to find that `eas build --local` solves both at once.

## Changes

- **`docs/DEPLOYMENT.adoc` → "Local production builds"** — restructured into two named cases:
  - **Case 1 (recommended)**: `eas build --local --profile production` runs EAS's pipeline on this machine, pulling the EAS upload keystore + remote versionCode at build time. Sideload the result with `adb install -r build-*.apk`. Strictly-greater versionCode + matching signature → upgrade-in-place, all data preserved.
  - **Case 2**: `npx expo run:android --variant release` produces a locally-signed APK that *cannot* upgrade an EAS install. Either uninstall (wipes data) or manually bump `android.versionCode` past the remote counter and accept the signature error.
- **`app.config.ts`** — bumps `android.versionCode` to `24` as a working floor for the case-2 path. EAS uses the remote counter regardless, so this doesn't affect the cloud pipeline.
- **`CLAUDE.md`** — short pointer to the recipe in the Development section, so future agents go straight to `eas build --local` for "deploy to my phone" tasks instead of re-hitting the install errors.

## Test plan

- [x] `eas build --local --profile production` runs to completion (currently in flight as of opening this PR — keystore fetched, native compiles in progress).
- [ ] Sideloaded APK upgrades the existing EAS-built install on my Pixel without an `INSTALL_FAILED_*` error.
- [ ] Wallets + Nostr login + message history all survive the upgrade.
- [x] AsciiDoc renders cleanly (no malformed table or unclosed source block).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -5,6 +5,12 @@
 - Use `npm start` (not `npx expo start`) — the start script includes `--dev-client` which is required for custom native modules
 - Native rebuild required after changing plugins or native modules: `npx expo run:android`
 
+## Deploying a release APK to a physical device
+
+- The user's primary device is a Pixel running an EAS-built production install. **`npx expo run:android --variant release` will not upgrade it** — locally-signed APK fails with `INSTALL_FAILED_UPDATE_INCOMPATIBLE` (different keystore than EAS) and `INSTALL_FAILED_VERSION_DOWNGRADE` (local versionCode resets to 1 every prebuild, while EAS's remote counter is in the high 20s).
+- To preserve the user's app data (wallets, Nostr login, message history), use `eas build --local --profile production --platform android --non-interactive` instead. This runs EAS's pipeline on this machine, fetches the EAS upload keystore so the signature matches, and uses the remote-incremented versionCode. Sideload with `adb -s <serial> install -r build-*.apk` (note: `adb` takes the serial, `expo run:android --device` takes the model name like `Pixel_8`).
+- Full recipe + rationale (case 1 vs case 2) lives in `docs/DEPLOYMENT.adoc` → "Local production builds". When a deploy fails for one of these reasons, update that section if anything's changed.
+
 ## Testing
 
 - E2E tests use Maestro and live in `tests/e2e/`
--- a/app.config.ts
+++ b/app.config.ts
@@ -113,6 +113,15 @@ export default ({ config }: ConfigContext): ExpoConfig => ({
     },
     predictiveBackGestureEnabled: false,
     package: getAndroidPackage(),
+    // Floor for the local-build versionCode. EAS production builds use
+    // their own remote autoIncrement counter (`eas.json` →
+    // `appVersionSource: "remote"`) and ignore this. Local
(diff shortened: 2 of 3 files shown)


BenGWeeks (commented):
## Self-review

### Strengths

- **The Case 1 vs Case 2 split is the right framing** — most readers want "upgrade in place over my EAS install without losing data" (case 1), and only a minority want the gradle-direct path (case 2). Naming each path makes the recommendation explicit.
- **The signature-mismatch + version-downgrade compounding** is correctly explained as TWO unrelated errors that stack — without that, a reader might fix one and conclude the path is broken when really both need to be addressed.
- **The `CLAUDE.md` pointer** points future agents straight to the recipe so this doesn't get re-discovered.
- **`versionCode: 24` floor** is correctly described as not affecting EAS's remote counter (`eas.json` has `appVersionSource: "remote"`).

### Concerns / nits

1. **Conflict with PR #242** — both PRs add adjacent sections to `docs/DEPLOYMENT.adoc`. Whichever lands second needs a manual rebase. Already called out in this PR's body, good.
2. **`versionCode: 24` is a moving target** — the next time someone runs case 2, EAS's remote counter will be higher than 24 again and the install will fail. Worth updating the case-2 recipe to say "check `eas build:list` first, set the in-repo `versionCode` to (latest-EAS) + 1" — actually the recipe already does say this, so this comment is just confirming it's there.
3. **Suggestion** — the case-1 recipe doesn't mention that `eas build --local` requires `eas-cli` installed locally + the user logged in (`eas whoami`). One-line prereq note would save a future reader an error.
4. **Sub-nit** — the `EXPO_NO_DOCTOR=1` env var in case 1 isn't strictly needed for the fix; documented elsewhere in the file already. Consistent with the existing Dogfooding section though, so fine to keep.

### Verification gap

The PR's test plan has "Sideloaded APK upgrades the existing EAS-built install on my Pixel without an `INSTALL_FAILED_*` error" unchecked. We did verify this end-to-end in this session (versionCode 25 installed cleanly, app data preserved). Would be worth checking the box before merge.

PR #242, opened 2026-04-27
docs(deployment): wireless adb recipe for sideloading when USB fails

BenGWeeks:
## Summary

Documents the workflow needed when the laptop won't see the phone over USB regardless of cable or port. Hit this today while installing a local prod build to my Pixel — none of the cables made `lsusb` show the device, so we fell back to wireless adb. Documenting before the gotchas evaporate.

Adds a new `===== Connecting the phone for sideload` subsection inside *Local Builds → Dogfooding*. Two parts:

1. **Diagnose USB before giving up.** `lsusb | grep -iE "google|18d1"` works without adb authorization — clean signal of cable / port health. If lsusb is empty even after re-plugging, it's physical (charge-only cable, lint in the USB-C socket, dead port), and further `adb` poking won't help. Switch to wireless.
2. **Full wireless-adb recipe** (Android 11+ — Pixel 8 confirmed). Step-by-step from enabling Wireless debugging through `adb pair` + `adb connect`, with the gotchas that cost us time today:

## Gotchas captured

- **The pairing port and the connection port are different.** Pair port appears only inside the *Pair device with pairing code* dialog; connection port is on the main *Wireless debugging* screen. Connecting to the pair port leaves you stuck on `offline`.
- **Pair codes are one-shot.** `error: protocol fault (couldn't read status message): Success` after `adb pair` doesn't mean the code is wrong — it means the code was burned by an earlier failed attempt. Generate codes on demand, not in advance.
- **mDNS auto-discovery is empty on Wi-Fi networks that block multicast.** `adb mdns services` returns nothing on guest networks and some routers. Manual pair + connect with explicit IPs still works; only the auto-discovery convenience is gone.

## Test plan

- [x] AsciiDoc renders cleanly (no malformed table or unclosed source block).
- [x] Followed the recipe end-to-end on this laptop + Pixel 8 to install a local prod APK after USB stayed dead through three cable swaps.

## Related

Touches the same file as #241 (which adds the version-downgrade + locally-signed-APK gotchas). The two PRs add disjoint sections, but if both reach merge they'll need a manual conflict resolution since they're adjacent in the file.

### DIFF
--- a/docs/DEPLOYMENT.adoc
+++ b/docs/DEPLOYMENT.adoc
@@ -106,6 +106,53 @@ adb -s <serial> shell dumpsys package com.lightningpiggy.app | \
   grep -E "versionCode|lastUpdateTime"
 ----
 
+===== Connecting the phone for sideload
+
+`adb` first needs to see the phone. USB is the happy path, but in practice the dog-eared cable or a stuck USB controller often defeats it — when that happens, fall back to wireless adb instead of debugging the cable. Both phone and laptop must be on the same Wi-Fi.
+
+*Diagnose USB before giving up:*
+
+[source,bash]
+----
+# Check USB enumeration at the kernel level — this works without
+# adb authorization, so it's a clean signal of cable / port health.
+# A working Pixel shows up as `Bus NNN Device NNN: ID 18d1:...`
+# (Google vendor ID 18d1).
+lsusb | grep -iE "google|18d1"
+
+# If lsusb is empty even after re-plugging, the laptop doesn't see
+# the phone at all — it's the cable, the port, or lint in the
+# phone's USB-C socket. Don't keep poking adb; switch to wireless.
+----
+
+*Wireless adb (Android 11+ — Pixel 8 supports it):*
+
+. *Phone:* Settings → System → Developer options → Wireless debugging → ON.
+. *Phone:* tap the *Wireless debugging* row (not just the toggle) to open the detail page.
+. The detail page shows an **IP address & Port** at the top — this is the **persistent connection port** (no code beside it). Note it.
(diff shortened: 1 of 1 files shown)


BenGWeeks (commented):
## Self-review

### Strengths

- **`lsusb | grep -iE "google|18d1"` as the diagnostic gate is excellent** — it's the cleanest split between "physical USB problem" and "adb authorization / debug-mode" issues, and it works without phone-side adb auth (so you can run it before the phone has accepted the host's RSA fingerprint).
- **The three gotchas are all real and were each costly today**: pair-port vs connect-port, one-shot pair codes, mDNS blocked on multicast-restricted Wi-Fi. Anyone hitting wireless adb for the first time will hit all three.
- **Same-Wi-Fi requirement** is mentioned ("Both phone and laptop must be on the same Wi-Fi"). Good.
- **The `connect-port` confusion** is the most common one based on stackoverflow / RN discord — calling it out by name with a clear description of where each port lives is the highest-value piece of the doc.

### Minor

- The recipe doesn't mention that wireless adb requires Android 11+. The Pixel 8 is fine, but a reader on an older device would hit a missing-menu-item dead end. One-line prereq.
- Worth adding what to do when wireless adb itself stops working mid-session (e.g. phone reboots, IP changes). Quick re-pair recipe? Actually `adb connect <ip:port>` should re-establish without a re-pair as long as the host's RSA cert is still trusted.
- The `error: protocol fault (couldn't read status message): Success` error message is verbatim correct. Future readers will be able to grep for it. Good.

### Conflict watch

The PR description correctly calls out the `docs/DEPLOYMENT.adoc` conflict with PR #241 — both PRs add adjacent sections. Whichever lands second needs a manual conflict resolve.

PR #244, opened 2026-04-27
perf(ui): stop 60x idle CPU + drop list-render keystroke loss (#243)

### LINKED ISSUE #243: perf: re-render storm causes dropped keystrokes, cursor jumps, and general lag
## Summary

Three symptoms reported today that all match the React Native "parent re-render storm" pattern — the same root cause is likely behind all of them.

## Symptoms

1. **Group Name input drops characters while typing.** Typing "Family" into the *Create Group* sheet's name field results in a partial string ("Famly", "Faily" etc.). The sheet uses `BottomSheetTextInput` (per the workaround in `docs/TROUBLESHOOTING.adoc` for plain `TextInput`-in-sheet keystroke loss), so the existing fix is insufficient against this stronger trigger.
2. **Seed Phrase cursor jumps when editing.** On *Add Hot Wallet → Step 1: Import Seed Phrase*, typing characters works fine, but **deleting/editing** characters causes the cursor to jump unpredictably.
3. **App generally laggy.** Background work appears to be slowing the UI thread — interactions feel non-instantaneous regardless of which screen is active.

## Likely root cause

A context provider (most likely `WalletContext` or `NostrContext`) is pushing state updates frequently enough to re-render the whole component tree under it. Every re-render of an open sheet:
- Invalidates `<TextInput>` cursor position → cursor jumps
- Drops keystrokes mid-IME-composition → "Famly" instead of "Family"
- Burns frames → general lag

## Already eliminated (no need to re-investigate)

| Suspect | Verdict |
|---|---|
| `WalletContext` `expectPayment` 1s `setInterval` (line 1297) | Only fires while Receive sheet is open with an active invoice. Not the cause of general lag or non-Receive-sheet typing issues. |
| `WalletContext` connection-check 30s `setInterval` (line 399) | Too slow to drop per-keystroke characters. |
| `WalletContext` active-wallet balance refresh 30s `setInterval` (line 1442) | Same — too slow. |
| `BottomSheetTextInput` swap | Already in place in `CreateGroupSheet.tsx`. The keystroke-drop bug exists *despite* the workaround, so the issue is upstream of the sheet input itself. |

## Diagnostic plan

1. **Profile a typing session.** With dev build + React DevTools' Profiler tab attached, type into the Group Name field and capture the flame chart. Look for components re-rendering >5× during the typing window.
2. **Inspect `NostrContext`** for `setContacts` / `setMessages` calls fired by relay events — likely the highest-frequency state pusher. Confirm whether incoming-event handlers debounce or fire per-event.
3. **Check the context Provider's `value` prop** for unstable object references (a fresh `{...}` on every render forces all consumers to re-render even when state hasn't actually changed). Should be `useMemo`-wrapped.
4. **Inspect `CreateGroupSheet.tsx`** — `sortedContacts.map(...)` (line 154) renders ~50 contact rows inline on every render of the sheet, each with an `<Image>`. Memoising the row component or switching to `<FlatList>` might be enough on its own to stop the keystroke drops in this specific sheet.

## Reproduction

* `expo start` → run on emulator or Pixel
* Sign in → Friends tab → Groups → tap "+" → tap *Group Name* → type "Family" reasonably quickly
* Symptom: input shows a partial string (drop rate seems to scale with the number of contacts in the friends list)

For the seed-phrase symptom: *Add Wallet → Hot Wallet → paste any seed → try to edit/delete characters* → cursor jumps.

## Out of scope

* Existing TROUBLESHOOTING entry for plain `TextInput`-in-sheet keystroke loss — covered, but insufficient.
* Local-prod-build and emulator-grey-screen issues from today (#240, #241, #242) — separate workflow concerns, unrelated to the runtime perf issue.

🤖 Filed by Claude Code based on real-time symptom reports during 2026-04-27 session.

BenGWeeks:
## Summary

Three compounding causes of the perf symptoms in #243 (dropped keystrokes, cursor jumps, general lag), fixed together because the relief from any one is bounded by the others.

### 1. Avatar GIFs were spinning 4 frame-decoder threads at >90% CPU

`expo-image` defaults to **autoplay** for animated GIF / WebP avatars. On a fresh AVD signed in as Big Piggy, `top -H` showed continuous frame decoding for every animated avatar in every mounted list row — dwarfing every other source of lag.

| Per-thread CPU | Before | After |
|---|---|---|
| `FrameDecoderExe` (× 4) | 92-100% each | not visible in top |
| `RenderThread` | 92.8% | <1% |
| **App total** | **451%** | **7.4%** |

Added `autoplay={false}` on four list-density avatar callsites: list rows (`ConversationRow`, `CreateGroupSheet`'s new `MemberRow`, `FriendPickerSheet`) and `GroupAvatar`. Shows the first frame only; tapping through to a profile sheet still shows the live animation, so the only thing suppressed is background-loop decoding for list rows.

### 2. `WalletContext` + `GroupsContext` leaked re-renders to every consumer

Both Providers passed an inline `value={{...}}` object literal — fresh identity on every render. Combined with `WalletContext`'s 30s connection-check + 30s balance-poll + 1s `expectPayment` tick, every ambient state update cascaded into a re-render of every `useWallet()` / `useGroups()` consumer in the tree, including any open `BottomSheetTextInput`, which dropped in-flight keystrokes and snapped cursor position to end.

Wrapped both `value` objects in `useMemo` with a full deps array. All exposed callbacks were already `useCallback`'d, so identities are stable and the memo bailout actually fires. Matches the existing `NostrContext` / `ThemeContext` pattern.

### 3. `CreateGroupSheet` contact picker was rendering 50 rows inline per keystroke

The `sortedContacts.map(c => <TouchableOpacity ... <Image ... />)` inline render meant every keystroke in the Group Name field re-built JSX for ~50 contact rows. Extracted `MemberRow` into a `React.memo` component and made `toggle` `useCallback`-stable so the memo bailout actually skips re-render when only `name` changed.

## What's still TODO (deferred to a follow-up)

The bigger win for picker / list density is **virtualization** — switching the inline `.map` to a `FlatList` so offscreen rows aren't even mounted. That's a bigger refactor (BottomSheetScrollView wrapper, sticky-header handling) and not blocking on the symptom relief delivered here. Filing a follow-up.

## Verification

* **Per-thread CPU** post-launch (above table) — 60× reduction in idle CPU.
* **Theoretically correct** for #2: textbook React perf — unstable Provider `value` → all consumers re-render → cascading re-renders interrupt the IME pipeline.
* **Manual:** typing in the Group Name input should now hold all keystrokes; cursor in the seed-phrase field should no longer jump on edits; navigation should feel instantaneous instead of laggy.

## Test plan

- [x] `npx prettier --check` clean on changed files
- [x] `npx eslint` clean on changed files
- [x] Idle-CPU measurement on a fresh AVD: 451% → 7.4%
- [ ] Manual on the dev build: type "Family" in Group Name field, all letters land
- [ ] Manual on the dev build: edit/delete characters in the Import Seed Phrase field, cursor stays put
- [ ] Manual on the dev build: navigation between Friends → Messages → Home feels responsive

Closes #243

### DIFF
--- a/src/components/CreateGroupSheet.tsx
+++ b/src/components/CreateGroupSheet.tsx
@@ -30,8 +30,10 @@ interface Props {
   onCreated?: (group: Group) => void;
 }
 
+type ContactWithLabel = NostrContact & { displayName: string };
+
 interface MemberRowProps {
-  contact: NostrContact;
+  contact: ContactWithLabel;
   isSelected: boolean;
   onToggle: (pubkey: string) => void;
   styles: ReturnType<typeof createStyles>;
@@ -45,11 +47,7 @@ interface MemberRowProps {
 // flood-render that dropped keystrokes mid-IME-composition. See #243.
 const MemberRow = React.memo<MemberRowProps>(
   ({ contact, isSelected, onToggle, styles, colors }) => {
-    const displayName =
-      contact.profile?.displayName ||
-      contact.profile?.name ||
-      contact.petname ||
-      contact.pubkey.slice(0, 12);
+    const { displayName } = contact;
     return (
       <TouchableOpacity
         style={styles.row}
@@ -155,12 +153,16 @@ const CreateGroupSheet: React.FC<Props> = ({ visible, onClose, onCreated }) => {
     });
   }, []);
 
-  const sortedContacts = useMemo(() => {
-    return [...contacts].sort((a, b) => {
-      const an = (a.profile?.displayName || a.profile?.name || a.petname || a.pubkey).toLowerCase();
-      const bn = (b.profile?.displayName || b.profile?.name || b.petname || b.pubkey).toLowerCase();
-      return an.localeCompare(bn);
-    });
+  // Compute `displayName` once per contact here rather than in MemberRow's
+  // render — saves the `||` chain firing on every keystroke-driven re-render
+  // when memo bailout doesn't apply. Sort uses the same key for consistency.
+  const sortedContacts = useMemo<ContactWithLabel[]>(() => {
+    const labeled: ContactWithLabel[] = contacts.map((c) => ({
+      ...c,
+      displayName: c.profile?.displayName || c.profile?.name || c.petname || c.pubkey.slice(0, 12),
+    }));
+    labeled.sort((a, b) => a.displayName.toLowerCase().localeCompare(b.displayName.toLowerCase()));
+    return labeled;
   }, [contacts]);
 
   const handleCreate = async () => {
--- a/src/components/CreateGroupSheet.tsx
+++ b/src/components/CreateGroupSheet.tsx
@@ -22,12 +22,89 @@ import type { Palette } from '../styles/palettes';
 import { useNostr } from '../contexts/NostrContext';
 import { useGroups } from '../contexts/GroupsContext';
 import type { Group } from '../types/groups';
+import type { NostrContact } from '../types/nostr';
 
 interface Props {
   visible: boolean;
   onClose: () => void;
   onCreated?: (group: Group) => void;
 }
+
+interface MemberRowProps {
+  contact: NostrContact;
+  isSelected: boolean;
+  onToggle: (pubkey: string) => void;
+  styles: ReturnType<typeof createStyles>;
+  colors: Palette;
+}
+
+// React.memo means rows that didn't change props (typing in the Group
+// Name field doesn't touch any of `contact`, `isSelected`, `onToggle`,
+// `styles`, `colors`) skip re-render entirely. With ~50 contacts each
+// rendering an Image, the inline `.map` was producing the
+// flood-render that dropped keystrokes mid-IME-composition. See #243.
+const MemberRow = React.memo<MemberRowProps>(
+  ({ contact, isSelected, onToggle, styles, colors }) => {
+    const displayName =
+      contact.profile?.displayName ||
+      contact.profile?.name ||
+      contact.petname ||
+      contact.pubkey.slice(0, 12);
+    return (
+      <TouchableOpacity
+        style={styles.row}
+        onPress={() => onToggle(contact.pubkey)}
+        accessibilityLabel={`${isSelected ? 'Deselect' : 'Select'} ${displayName}`}
+        testID={`member-row-${contact.pubkey.slice(0, 12)}`}
+      >
+        <View style={styles.avatar}>
+          {contact.profile?.picture ? (
+            <Image
+              source={{ uri: contact.profile.picture }}
+              style={styles.avatarImage}
+              cachePolicy="disk"
+              // First frame only for animated WebP / GIF avatars.
+              // Without this, expo-image spawns a FrameDecoderExe
+              // thread per animated avatar and decodes every frame on
+              // a continuous loop. Saw 4 threads at >90% CPU each on a
+              // fresh AVD with ~50 contacts in the picker. Tapping
+              // through to a profile sheet still shows the live
+              // animation. See #243.
+              autoplay={false}
+            />
+          ) : (
+            <Svg width={20} height={20} viewBox="0 0 24 24" fill="none">
+              <Circle cx="12" cy="8" r="4" fill={colors.textSupplementary} />
+              <Path
+                d="M4 20c0-3.314 3.582-6 8-6s8 2.686 8 6"
+                stroke={colors.textSupplementary}
+                strokeWidth={2}
+                strokeLinecap="round"
+              />
+            </Svg>
+          )}
+        </View>
+        <Text style={styles.rowName} numberOfLines={1}>
+          {displayName}
+        </Text>
+        <View style={[styles.checkbox, isSelected && styles.checkboxActive]}>
+          {isSelected && (
+            <Svg width={14} height={14} viewBox="0 0 24 24" fill="none">
+              <Path
+                d="M20 6 9 17l-5-5"
+                stroke={colors.white}
+                strokeWidth={3}
+                strokeLinecap="round"
+                strokeLinejoin="round"
+              />
+            </Svg>
+          )}
+        </View>
+      </TouchableOpacity>
+    );
+  },
+);
+MemberRow.displayName = 'MemberRow';
 
 const CreateGroupSheet: React.FC<Props> = ({ visible, onClose, onCreated }) => {
   const colors = useThemeColors();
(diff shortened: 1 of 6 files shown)


src/contexts/WalletContext.tsx
```diff
@@ -1466,46 +1474,85 @@ export const WalletProvider: React.FC<{ children: React.ReactNode }> = ({ childr
...
+      reorderWallet,
+      setActiveWallet,
+      refreshActiveBalance,
+      completeOnboarding,
+      makeInvoice,
+      payInvoice,
+      makeInvoiceForWallet,
+      payInvoiceForWallet,
+      refreshBalanceForWallet,
+      fetchTransactionsForWallet,
```
BenGWeeks, line 1546: **Nit — redundant deps**: `activeWallet`, `hasWallets`, `isConnected`, `balance`, `walletAlias` are all derived from `wallets` and `activeWalletId` (or `activeWallet`, which itself is derived). When `wallets` changes, every derived value gets a new reference, so the useMemo invalidates already. Including the derived values in the deps adds extra reference-equality checks per render but no correctness benefit.

Not blocking — the value object is the same shape either way and memo correctness is preserved. If you do trim, the minimal stable deps are: `[wallets, activeWalletId, isOnboarded, isLoading, currency, btcPrice, lastIncomingPayment, ...all callbacks]`.

src/components/CreateGroupSheet.tsx
```diff
@@ -22,13 +22,90 @@ import type { Palette } from '../styles/palettes';
...
+        <Text style={styles.rowName} numberOfLines={1}>
+          {displayName}
+        </Text>
+        <View style={[styles.checkbox, isSelected && styles.checkboxActive]}>
+          {isSelected && (
+            <Svg width={14} height={14} viewBox="0 0 24 24" fill="none">
+              <Path
+                d="M20 6 9 17l-5-5"
+                stroke={colors.white}
+                strokeWidth={3}
```
BenGWeeks, line 94: **Suggestion — `displayName` recomputed per row render**: each `MemberRow` render runs the `||` chain (~5 ops) to derive `displayName`. Cheap, but the row's `accessibilityLabel` and the visible `<Text>{displayName}</Text>` both consume it. Consider hoisting to the parent so it ships in the contact data once:

```tsx
// In CreateGroupSheet (where sortedContacts is built)
const contactsWithLabels = useMemo(
  () => sortedContacts.map(c => ({
    ...c,
    displayName: c.profile?.displayName || c.profile?.name || c.petname || c.pubkey.slice(0, 12),
  })),
  [sortedContacts],
);
```

Then `MemberRow` reads `contact.displayName` and the derivation only re-runs when the contact list itself changes. Marginal win for 50 rows; ignore if not bothering.

src/components/GroupAvatar.tsx
```diff
@@ -110,6 +110,8 @@ const SingleAvatar: React.FC<SingleAvatarProps> = ({ picture, colors, innerSize
           style={{ width: innerSize, height: innerSize, borderRadius: innerSize / 2 }}
           cachePolicy="disk"
           onError={() => setErrored(true)}
+          // First frame only — see #243.
+          autoplay={false}
         />
```
BenGWeeks, line 123: **Verify** — `GroupAvatar` is used in `GroupRow.tsx:84` (the Groups list). That's a list-density context, so `autoplay={false}` is correct here.

If `GroupAvatar` ever ends up being reused in a *single-avatar* context (e.g. the group conversation header) where animation IS wanted, this needs to become a prop (`autoplayAvatar?: boolean`) rather than hard-coded. Worth a comment on the component signature noting the assumption.

BenGWeeks (commented):
## Self-review (no Copilot review yet — will hold for that)

### Strengths

- **Measurement evidence is excellent**: 451% → 7.4% CPU is unambiguous and the per-thread `top -H` breakdown isolates `FrameDecoderExe` cleanly. This is a well-evidenced perf fix.
- **Layered fix is the right shape**: the `autoplay={false}` is the dominant CPU win, the context `useMemo` is the textbook React fix that the keystroke-loss symptom would *also* require even if the avatar issue didn't exist, and the `MemberRow` extraction is the local sheet-specific cherry-on-top.
- **`React.memo` + stable `toggle` `useCallback` is correctly aligned**: empty `[]` deps is right because `setSelected(prev => …)` uses the functional updater. The memo bailout will actually fire.
- **Comments explain *why*, not what**: each `autoplay={false}` site has a one-line `See #243` and one site has the full root-cause explanation. Future readers won't have to guess.
- **Good restraint** on filing FlatList virtualization as a follow-up rather than scope-creeping it into this PR.

### Minor concerns / nits (none blocking)

See inline.

### Out of scope but worth flagging

- `FlatList` virtualization is the *real* fix for the list-density case — `autoplay={false}` is a UX regression (animated avatars no longer animate in lists at all, even when visible). The proper follow-up is to switch the inline `.map`s to `<FlatList>` and reinstate `autoplay={true}` (the default). With virtualization, only ~10 visible rows are mounted, so frame decoding stays bounded. **File a follow-up issue** before this merges so it doesn't get lost — the user explicitly raised this concern ("can we not just animate the ones in view?") during the investigation.
- `MissionDetailScreen.tsx`, `account/ProfileScreen.tsx`, `SendSheet.tsx` etc. still have `<Image source={{ uri }}>` without `autoplay={false}` — but those are *single*-avatar contexts where animation may genuinely be wanted, so leaving them as-is is correct. Worth a comment in the PR body confirming this was deliberate, since a future reviewer might assume the sweep was incomplete.

### Verification questions for the author

- [ ] Confirm the emulator measurement was on the dev build *with* this PR's changes loaded via Metro (not a stale bundle from before the edits).
- [ ] Manual test on the live dev build: type "Family" in Group Name input, edit characters in the Import Seed Phrase field. Both should now hold all keystrokes / not jump cursor. PR description has this as an unchecked manual test plan item.

src/contexts/WalletContext.tsx
BenGWeeks: Trimmed in 7e35aa4 — removed `activeWallet`, `hasWallets`, `isConnected`, `balance`, `walletAlias` from the dep array; left an explanatory comment + `eslint-disable-next-line react-hooks/exhaustive-deps` since the rule can't see the derivation chain.

src/components/CreateGroupSheet.tsx
BenGWeeks: Hoisted in 7e35aa4 — `sortedContacts` now produces `ContactWithLabel[]` with `displayName` precomputed once per contact-list change. `MemberRow` reads `contact.displayName` directly. Folded into the existing memo (rather than a second one) and the sort uses the same key for consistency.

src/components/GroupAvatar.tsx
BenGWeeks: Done in 7e35aa4 — added a comment on the `GroupAvatar` component signature noting the list-density assumption and that any future single-avatar consumer (e.g. a group-conversation header) should promote `autoplay` to an `autoplayAvatar?: boolean` prop rather than flipping the literal.

src/components/FriendPickerSheet.tsx
```diff
@@ -196,6 +196,8 @@ const FriendPickerSheet: React.FC<Props> = ({
               source={{ uri: item.picture }}
               style={[StyleSheet.absoluteFillObject, styles.avatarImage]}
               cachePolicy="disk"
+              // First frame only — see #243.
+              autoplay={false}
```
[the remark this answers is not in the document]
BenGWeeks: Fixed in 7e35aa4 — added `styles` and `colors` to the `useCallback` dep array with a short comment explaining that without them the callback closes over a stale theme after a light/dark swap.

Issue #245, opened 2026-05-02
perf(ui): Friends + Messages tabs jank — 14–18% legacy jank, 5s GPU stalls from cold image decode


BenGWeeks:
## Summary

Both the Friends and Messages tabs are noticeably sluggish on cold open and during scroll. AVD measurements (dev variant on `main` @ `cb6c9f9`):

| Metric | Friends | Messages | Healthy |
|---|---|---|---|
| Janky frames | 9.7% | 15.3% | <5% |
| Janky frames (legacy) | **14.3%** | **18.6%** | <10% |
| 50th frame | 16ms | 16ms | 16ms (60fps) |
| 90th frame | 31ms | 31ms | <16ms |
| 99th frame | 450ms | 300ms | <33ms |
| **GPU 99th** | **4950ms** | **4950ms** | <16ms |

Both tabs hit 4.95-second GPU stalls — same root cause across both. Caveat: dev-mode + AVD; release-on-device deltas will be smaller in absolute terms but the relative shape will hold.

Reproduce: `./scripts/perf-friends-tab.sh main` and `./scripts/perf-messages-tab.sh main` against `emulator-5554`.

### HIGH — likely root cause of the 5s GPU stalls

1. **`friends-bg.png` is 754 KB rendered via stock RN `Image` on every tab mount** (`FriendsScreen.tsx:296-300`, `MessagesScreen.tsx:285-289`). No `expo-image`, no cache hint, no downsizing. Decoded on the UI thread before first paint — exactly the cold-decode profile we see in `gfxinfo`. PNG is 600x600 8-bit RGB; lossless but oversized for an avatar background.

### HIGH — render-storm patterns (jank %)

2. **`MessagesScreen.tsx:262-281`** — `renderItem` depends on `contactPictureMap`, which churns on every relay update (every kind-0 drip). That invalidates the `useCallback` → busts `React.memo` on every `ConversationRow`/`GroupRow` → 50-row re-render per profile drip-in.
3. **`GroupRow.tsx:25,47`** — linear `contacts.find(...)` per row, per render. `senderName` recompute is O(rows × contacts) on every \`contacts\` array identity change. The \`contactPictureMap\` pattern was meant to kill exactly this — but `GroupRow` doesn't use the map for name lookup.
[...]

PR #252, opened 2026-05-03
perf(ui): skeletons + deferred mount across all 4 tabs + 2 sheets

BenGWeeks:
## Summary

Drives perceived perf from "blank flash → JS chugs → content appears" to "skeleton appears immediately → animation completes smoothly → real content takes over". Targets the cold-paint surfaces on every bottom tab + the two heavy sheets that make up the (+) "new chat" path Ben specifically flagged as slow.

Foundation + apply, in one PR. Layer 3 (avatar prefetch) deliberately deferred to a follow-up so the perf delta from skeletons + defer is measurable in isolation.

## Why

Current AVD measurements showed cold-paint modern jank in the **30–60% range** across these surfaces. Root cause isn't bad code — it's that the heavy work (50 contact rows, 50 avatar decodes, sheet measurement, list-build) fires **synchronously while the UI is animating**. The user taps the FAB; the JS thread immediately starts allocating + queuing image decodes; the bottom-sheet's open animation contends with all of it.

Stop doing the heavy work during the animation. Show a skeleton in the meantime. Real content mounts after \`InteractionManager.runAfterInteractions\` so the animation completes uninterrupted. See \`/home/benw/.claude/plans/ok-how-do-we-noble-gem.md\` for the full strategy doc the user approved.

### 1. New foundation: \`<SkeletonRow>\` + \`<SkeletonList>\`

\`src/components/SkeletonRow.tsx\` (164 lines, no new deps):

- Configurable shape: avatar size, line count (1 or 2), height
- Shimmer effect via \`expo-linear-gradient\` + \`react-native-reanimated\` (both already in deps)
- Animation runs on the UI thread (via reanimated worklet) — no JS-thread cost per frame
- Theme-aware (\`useThemeColors()\`)
- \`SkeletonList\` convenience wrapper for stacking N rows

### 2. Apply across all 4 bottom tabs + 2 sheets

| Surface | What we did | Trigger |
|---|---|---|
| **Home** (\`HomeScreen.tsx\`) | 5 transaction-row skeletons replace the centred \`ActivityIndicator\` | While \`loadingTransactions === true\` |
| **Messages** (\`MessagesScreen.tsx\`) | 6 conversation-row skeletons | Until \`dmInbox\`/\`wallets\` hydrate or first \`InteractionManager\` tick |
| **Learn** | **None — out of scope per plan** (data is local + sub-second cold paint already) | n/a |
| **Friends** (\`FriendsScreen.tsx\`) | 10 contact-row skeletons | Until \`contacts\`/\`phoneContacts\` hydrate or first \`InteractionManager\` tick |
| **FriendPickerSheet** | 8 row skeletons + defer real list mount past sheet animation | \`useEffect([visible])\` → InteractionManager.runAfterInteractions → \`setListReady(true)\` |
| **CreateGroupSheet** | Same defer pattern | Same |

### 3. Row-height constants exported (for skeleton sizing without layout shift)

| File | New export |
|---|---|
| \`ConversationRow.tsx\` | \`CONVERSATION_ROW_HEIGHT = 72\` |
| \`GroupRow.tsx\` | \`GROUP_ROW_HEIGHT = 72\` |
| \`CreateGroupSheet.tsx\` | \`MEMBER_ROW_HEIGHT = 60\` |
| \`TransactionList.tsx\` | \`TRANSACTION_ROW_HEIGHT = 60\` |

(\`CONTACT_LIST_ITEM_HEIGHT = 72\` was already exported from \`ContactListItem.tsx\` — same pattern, just propagated to the others.)

**Why heights matter:** if the skeleton row is taller/shorter than the real row, replacing it produces visible layout shift = its own jank. Exporting the constant lets each surface size its skeleton against the canonical row height.

## Trade-offs accepted

- **Slight delay before real content** (skeleton shows ~250–400ms during the animation). Strictly worse than instant if there's nothing competing for the thread, but the animation we're guarding has its own duration anyway — the user just sees the skeleton instead of frame drops.
- **One additional component to maintain** (SkeletonRow, single file, no external lib).

## Test plan

- [x] \`npx tsc --noEmit\` clean
- [x] \`npx eslint <changed files>\` clean (6 pre-existing warnings unrelated to this change)
- [x] \`npx prettier --check\` clean
- [ ] **AVD perf re-measurement** with \`scripts/perf-friends-tab.sh\` from PR #246 + the inline FAB→FriendPicker open harness from this session — 5 cold samples each on this branch vs main. Success bar: modern jank during sheet-open phase < 10% (currently 34%).
- [ ] **Visual on AVD via Metro reload:**
  - Tap (+) on Messages → sheet animates up with skeleton rows visible → real list replaces skeletons after ~250ms with NO visible layout shift
  - Cold-launch app → tap Friends tab → skeleton rows visible while contacts hydrate → real rows replace cleanly
  - Cold-launch app → Home tab → skeleton transaction rows visible while wallet syncs → real transactions replace cleanly
- [ ] **Pixel release-build check** (combined APK with PR #246 + #250 + this branch). Subjective bar: "I don't see jank during sheet open."
- [ ] **Maestro regression sweep:** \`maestro test tests/e2e/test-friend-picker-alphabet.yaml\`, \`test-messages-search.yaml\`, \`test-3way-group-create-as-big.yaml\`. May need \`extendedWaitUntil\` on tests that assumed instant list visibility.

## Out of scope (follow-up issues)

- **Avatar prefetch on Messages tab focus** — pre-warm \`expo-image\` disk cache so cold sheet open hits warm bitmaps. Risk: prefetching all 50 avatars could spike network + memory; needs throttle + size cap. Separate PR.
- **\`useTransition\` on filter switches** — FriendsScreen filter chips, Messages window toggle. Small standalone PR.
- **Lazy tab-mount + memo audit** — broader work, separate scope.
- **Conversation message-bubble skeletons** — existing \`ActivityIndicator\` is adequate; revisit if subjective slowness on real device persists.
- **Standardising "Loading messages…" label vs spinner-only** — UX polish, separate concern.

## Related

- Issue #245 — umbrella perf issue (this PR partially addresses; doesn't close).
- PR #246 — perf-pass on caching + sort keys + render-storm patterns. Independent of this PR; both can land.
- PR #250 — group-conversation keyboard fix. Unrelated, can land independently.

### DIFF
--- a/src/components/ConversationRow.tsx
+++ b/src/components/ConversationRow.tsx
@@ -7,6 +7,14 @@ import type { Palette } from '../styles/palettes';
 import type { ConversationSummary } from '../utils/conversationSummaries';
 import { conversationPreview, formatConversationTimestamp } from '../utils/conversationSummaries';
 
+/**
+ * Total row height in dp: avatar 48 + paddingVertical 12 × 2 = 72.
+ * Exported so SkeletonRow can match the height exactly — mismatched
+ * heights produce layout shift the moment the skeleton is replaced.
+ * If you change paddingVertical or the avatar size below, update this too.
+ */
+export const CONVERSATION_ROW_HEIGHT = 72;
+
 interface Props {
   summary: ConversationSummary;
   onPress?: () => void;
--- a/src/components/CreateGroupSheet.tsx
+++ b/src/components/CreateGroupSheet.tsx
@@ -7,6 +7,7 @@ import {
   ActivityIndicator,
   BackHandler,
   Alert,
+  InteractionManager,
 } from 'react-native';
 import { Image } from 'expo-image';
 import Svg, { Path, Circle } from 'react-native-svg';
@@ -17,6 +18,7 @@ import {
   BottomSheetScrollView,
   BottomSheetTextInput,
 } from '@gorhom/bottom-sheet';
+import { SkeletonList } from './SkeletonRow';
 import { useThemeColors } from '../contexts/ThemeContext';
 import type { Palette } from '../styles/palettes';
 import { useNostr } from '../contexts/NostrContext';
@@ -31,6 +33,13 @@ interface Props {
 }
 
 type ContactWithLabel = NostrContact & { displayName: string };
+
+/**
+ * Total row height in dp: avatar 40 + paddingVertical 10 × 2 = 60.
+ * Exported so SkeletonRow can match the height exactly while the real
+ * member list is being built off the JS thread post-animation.
+ */
+export const MEMBER_ROW_HEIGHT = 60;
 
 interface MemberRowProps {
   contact: ContactWithLabel;
@@ -123,6 +132,22 @@ const CreateGroupSheet: React.FC<Props> = ({ visible, onClose, onCreated }) => {
     } else {
       sheetRef.current?.dismiss();
     }
+  }, [visible]);
+
+  // Defer the heavy member-list mount until after the sheet's open
+  // animation completes — same pattern as FriendPickerSheet. With ~50
+  // contacts each rendering an avatar Image, mounting them DURING the
+  // animation produces visible jank. Skeleton fills the area in the
+  // meantime; the real list mounts on the next interaction tick.
+  // See plan in #245 follow-up.
+  const [listReady, setListReady] = useState(false);
+  useEffect(() => {
+    if (!visible) {
+      setListReady(false);
+      return;
+    }
+    const handle = InteractionManager.runAfterInteractions(() => setListReady(true));
+    return () => handle.cancel();
   }, [visible]);
 
   useEffect(() => {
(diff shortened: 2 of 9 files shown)


BenGWeeks:
Closing — pivoting to local-first persistence per the WhatsApp-style approach.

After re-measuring on AVD (5 cold samples) the skeleton-with-correct-hydration version of this PR ended up at **65% modern jank** during sheet open — worse than baseline (60%) and worse than the animated-skeleton version (40%). Reading the histograms it's clear what happened: by holding the skeleton until \`dmInbox\` truly loads (using \`dmInboxLoading\`), we eliminated the "No conversations yet" flash but exposed the underlying 3-5s relay+decrypt work as measured frames. **Those janky frames were always there; the skeleton just stopped hiding them behind the empty-state UI.**

This proved Ben's point: **skeletons treat the symptom.** The real reason WhatsApp loads instantly without skeletons is that messages are already in local SQLite when the app opens — there's nothing to load.

Switching to: persist the decrypted NIP-17 DM inbox + group summaries to AsyncStorage on every change, rehydrate from storage on cold start before any UI renders. New PR coming. The skeleton component itself will be discarded.

The plan file at \`/home/benw/.claude/plans/ok-how-do-we-noble-gem.md\` captured the strategic discovery.

Branch + diff stays in the closed-PR record for reference.

PR #290, opened 2026-05-03
feat(onboarding): auto-provision CoinOS managed Lightning wallet (#287)

### LINKED ISSUE #287: feat(onboarding): auto-provision a CoinOS managed wallet + NWC for new users
## Summary

Let Lightning Piggy Mobile **create a CoinOS managed Lightning wallet** for new users as part of onboarding, returning a NIP-47 NWC connection string the app can use immediately. When the user later wants self-custody (their own LNbits / hardware NWC server), the app helps migrate funds out.

CoinOS = the "training wheels" wallet so a new user has a working Lightning experience from minute one without going through the LNbits / NWC setup walkthrough.

## Why CoinOS

Per [API research](https://github.com/coinos/coinos-server/blob/master/routes/users.ts):

- **Public, no-captcha account creation:** `POST /register { user: { username, password, ... } }` → returns `{ token: <JWT>, sk: <nostr secret> }`. The Google reCAPTCHA gate is on `/login` and `/nostrAuth` only — `/register` is wide open per the source.
- **NWC string available:** with the JWT, `POST /app` creates a connection; `GET /apps` returns objects whose `nwc` field is `nostr+walletconnect://${serverPubkey}?relay=${relay}&secret=${app.secret}&lud16=${user}@coinos.io`.
- **Self-hostable:** AGPL-3.0, Dockerized — supports the "graduate to self-custody" story (a sovereignty-minded user can spin up their own CoinOS, point LP at it, and migrate funds).
- **Already cosmetically integrated:** `src/themes/cards/coinos.ts` exists; LP renders a CoinOS-branded card variant. The runtime API client is missing.

## Proposed flow

1. New-user onboarding: tap "Create Lightning Wallet" → LP auto-generates a username (suggest `${nostr_npub_prefix}_lp`) + random strong password.
2. LP calls `POST https://coinos.io/register`, captures the returned JWT.
3. LP calls `POST https://coinos.io/app` (with JWT) + `GET https://coinos.io/apps`, captures the `nwc` field.
4. LP stores the NWC string in WalletContext (existing NWC wallet plumbing handles it from here).
5. LP shows a "Recovery info" screen on first paint: the username + password + NWC string + a one-line note "If you lose this device, restore by signing in to coinos.io with this username + password."

## Migration to self-custody (later PR)

When the user has a self-custody NWC server set up:
- LP shows a "Migrate funds" CTA on the CoinOS wallet card
- Drains the CoinOS balance via internal transfer to the new wallet (using existing TransferSheet plumbing)
- Disconnects the CoinOS NWC connection (revoke + remove from WalletContext)

## Acceptance criteria

- [ ] **ToS courtesy check** with CoinOS maintainers — programmatic mass registration from a mobile client is acceptable under their ToS. The lack of captcha on `/register` suggests yes, but worth a DM.
- [ ] Service file (`src/services/coinosService.ts` or similar) implementing register + app-create + nwc-fetch.
- [ ] Onboarding UI hook (probably in `IntroScreen` flow or a new `CreateLightningWalletScreen`).
- [ ] Recovery info screen — the user MUST be shown + asked to confirm they've backed up the username + password before proceeding. Otherwise a device wipe = funds lost.
- [ ] Wallet appears in `WalletContext`, payable / receivable like any other NWC wallet.

## Out of scope (file as follow-ups)

- The migrate-funds-out flow (separate PR — depends on this landing first).
- Showing a CoinOS-self-hosted endpoint picker in the create flow (let users point at their own instance immediately rather than coinos.com). Optional polish.
- Using a different default LSP (Voltage, Alby, Phoenix-API) — CoinOS is the recommended default given AGPL + the open `/register` endpoint, but the abstraction should let alternates plug in.

## Sources

- [coinos-server routes/users.ts](https://github.com/coinos/coinos-server/blob/master/routes/users.ts) — register + NWC field construction
- [coinos-server index.ts](https://github.com/coinos/coinos-server/blob/master/index.ts) — route registration
- [NIP-47 spec](https://nips.nostr.com/47)

BenGWeeks:
## Summary

- New "Create a Lightning Wallet" tile in the Add Wallet wizard auto-registers a managed wallet on CoinOS, mints a NIP-47 connection via `POST /app` + `GET /apps`, and feeds the resulting `nwc` string into the existing `addNwcWallet` plumbing — fresh users get working Lightning send/receive without manual NWC setup.
- New `src/services/coinosService.ts` is the only network surface: `registerCoinosUser`, `createCoinosNwcConnection`, `listCoinosApps`, `probeCoinosInstance`, plus username/password generators. Errors are mapped to a structured `CoinosError` (`code`: `username_taken | rate_limited | service_down | network | timeout | auth | invalid_input | unknown`) so the UI can render specific copy.
- New mandatory recovery-info sheet (`CoinosRecoverySheet`) shows the username, password, server URL, and full NWC string. In post-create mode the sheet is non-dismissable until the user taps **I've saved this somewhere**; in re-display mode (Wallet Settings → "View recovery info") it's freely closable. Username/password persist in SecureStore (`AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY`) so the user can come back to them; the password is never held in JS state outside an active sheet.
- `WalletSettingsSheet` for managed wallets gains **View recovery info** (re-shows the same sheet) and a greyed-out **Migrate to self-custody** row hinting at the follow-up issue. Wallet removal also drops the recovery record so we don't leak credentials for a wallet the user just removed.
- An Advanced expander on the create screen lets the user point at a self-hosted CoinOS instance — validated via `GET /health` before we commit, so a typo can't strand a half-registered account.

## Custody disclosure

This PR adds a path to a **custodial** wallet. The custody screen quotes the trust model verbatim per the issue:

> Your funds will be held by CoinOS, not by you. Suitable for testing or small amounts — not life savings. You can move to self-custody whenever you're ready.

CoinOS = Asoka Wong's single-LND-node operation under AGPL-3.0. Hot wallet, no vault, no multisig — fine for $5–100 onboarding / training-wheels use, NOT appropriate for life savings. The flow exists to give a brand-new user working Lightning in 10 seconds without facing the LNbits / NWC walkthrough; the long-term path is the (future) "Migrate to self-custody" CTA, which this PR scaffolds but does not implement.

The flow never auto-triggers — only an explicit user tap on "Create a Lightning Wallet" in the Add Wallet wizard fires the registration. We never silently provision an account on launch. Per the issue, the courtesy ToS DM to CoinOS maintainers (the `/register` endpoint is wide open per upstream `routes/users.ts`, but worth confirming bulk programmatic use is acceptable) is a precondition for un-drafting this PR.

## API client design

| Function | Purpose |
| --- | --- |
| `registerCoinosUser({ baseUrl, username, password })` | `POST /register` with `{ user: { username, password } }`. Returns `{ token, sk?, pubkey?, username }`. Validates `username` against `^[a-z0-9_]{3,32}$` and `password.length >= 12` before any network call. |
| `createCoinosNwcConnection({ baseUrl, token, name })` | Generates a 32-byte secret client-side, derives the pubkey via `nostr-tools` `getPublicKey(hexToBytes(secret))`, posts both to `/app`, then resolves the assembled NIP-47 connection string from `/apps` (the upstream `POST /app` handler returns `{}`; the `nwc` field is built only when reading the apps list). |
| `listCoinosApps({ baseUrl, token })` | `GET /apps` with the bearer JWT. Drops malformed entries silently. |
| `probeCoinosInstance(baseUrl)` | `GET /health` with an 8 s timeout. Used by the self-hosted picker. |
| `suggestUsername()` | `lp_<8 hex>` — 32 bits of CSPRNG entropy in the user-visible portion. |
| `generateStrongPassword()` | 30 bytes of CSPRNG → base64url, ≥40 chars (≥240 bits). |

Standalone (no `fetchWithTimeout` import — `nostrService` doesn't export one today; the helper is internal to the service file). All requests carry a 30 s deadline backed by `AbortController`.

`coinosService.test.ts` mocks `global.fetch` to cover validation, request shape, error classification (400/401/429/500 + thrown TypeError), and the secret-generation invariants. Does NOT call coinos.io at any point.

## Recovery-info screen design

- Brand-pink shield-alert bubble + "Save your recovery info" headline; subtitle quotes the host (`coinos.io` or self-hosted) so the user knows where to log in.
- Four labelled rows — Server, Username, Password, NWC connection — each with a copy-to-clipboard button on the right. Password + NWC use a monospace font; the NWC row truncates to 3 lines so the full string is reachable but doesn't dominate the screen.
- Below the rows: a transient "X copied to clipboard" confirmation, then the primary action ("I've saved this somewhere" in post-create mode, "Done" in re-display mode).
- Post-create mode: `enablePanDownToClose={false}` + `pressBehavior="none"` on the backdrop; the Android back gesture and a backdrop tap are no-ops. The user MUST press the primary button.

## Test plan

- [ ] `npx tsc --noEmit` passes (verified locally — only the pre-existing `tests/unit` Jest-types errors that affect main).
- [ ] `npx eslint` on changed files passes — only `props: any` warnings on `renderBackdrop` callbacks, which match every existing sheet in this codebase.
- [ ] `npx prettier --check` passes.
- [ ] Unit tests: `npx jest src/services/coinosService.test.ts` — coverage for input validation, request shape, error mapping, /apps parsing, instance probe.
- [ ] Manual: emulator + Pixel — Add Wallet → Create a Lightning Wallet → custody screen → spinner → recovery sheet → tap "I've saved this somewhere" → land on Home with the new wallet selected as active. Confirm balance loads + a small test invoice can be paid.
- [ ] Manual: Wallet Settings → "View recovery info" re-shows the same sheet without the lock. Greyed-out "Migrate to self-custody" is non-interactive.
- [ ] Manual: error paths — invalid self-hosted URL surfaces "Could not reach that CoinOS instance"; offline shows "Couldn't reach CoinOS"; hammering the create button quickly doesn't double-register.
- [ ] Manual: remove the wallet → confirm the SecureStore recovery key is dropped (cold-start the app, settings sheet should not show "View recovery info" for the next wallet at the same id).

## Pre-undraft checklist (per issue acceptance criteria)

- [ ] Courtesy DM to CoinOS maintainers (Asoka Wong) confirming programmatic registration from a mobile client is acceptable under their ToS.
- [ ] Inline screenshots — custody-warning, recovery-info, and post-create Home — captured on a fresh emulator install. (Skipped during draft to avoid restarting the running Metro on the main checkout; will be added before un-drafting.)
- [ ] File a follow-up issue for the "Migrate to self-custody" CTA (drain the CoinOS balance via the existing TransferSheet plumbing, then revoke the NWC connection).

Closes #287


BenGWeeks:
Screenshots needed.

PR #319, opened 2026-05-04
feat(messages): swipe-left to delete a message locally (#128)

### LINKED ISSUE #128: Swipe-left to delete a message in the conversation view
## Problem

Messages in the conversation view have no delete / unsend / hide affordance. Users who mistype, share the wrong invoice, or change their mind after sending a location have no way to remove it from their own chat history.

## Proposed

Add a swipe-left gesture on any message bubble (or invoice / location / contact card) in \`src/screens/ConversationScreen.tsx\` that reveals a red **Delete** action. Tap confirms and removes the message from the local view.

### Local delete only
Nostr events, once published, are immutable. A true cross-relay delete requires a NIP-09 deletion event (kind 5) referencing the original event's id — which relays and clients honour optionally, not universally. So \"swipe to delete\" should be scoped to **local-only hiding** first, with an optional later enhancement to also publish the NIP-09 deletion for clients that respect it.

### Persist the hidden set
Track deleted message ids in AsyncStorage (e.g. \`conversation_deleted_<pubkey>\`) so the hide survives relaunch. \`fetchConversation\` filters the returned events against this set before handing to the render path.

### What about outgoing vs. incoming?
- **Outgoing (fromMe):** full delete semantics — hide locally + optionally publish NIP-09.
- **Incoming (fromThem):** \"Hide for me\" only, never a publish. Relabel the swipe action to \"Hide\" for incoming to set expectations.

## Implementation sketch

1. Wrap each message row in a \`react-native-gesture-handler\` \`Swipeable\` (or the existing composition the project uses for swipe actions).
2. \`renderRightActions\` returns a red full-height button labeled Delete / Hide.
3. \`onSwipeableWillOpen\` or the button's onPress calls \`handleDeleteMessage(item.id)\`.
4. \`handleDeleteMessage\` appends to the hidden set, writes it back to AsyncStorage, and filters the render path.
5. Optional follow-up: for outgoing, also call a new \`publishDeletion(eventId)\` in nostrService that emits a kind-5 event tagging the original.

## Related / prior art

- Inverted \`FlatList\` per PR #124 means \"left\" in the inverted coordinate space is still \"left\" visually — no axis flip needed for the gesture.
- CLAUDE.md bans coordinate taps in Maestro, so the e2e test for this feature will need an \`accessibilityLabel\` on the delete button rather than a swipe gesture (\`tapOn\` by id after the swipe has been staged in a setup step, or a deep-link-seed approach similar to the NWC paste workaround).

## Notes

Surfaced on 2026-04-21 during the share-contact PR (#121) merge.

BenGWeeks:
## Summary

- Wraps each `MessageBubble` in a `react-native-gesture-handler` `ReanimatedSwipeable` so users can swipe-left to reveal a Delete (outgoing) / Hide (incoming) action — same affordance as WhatsApp / iMessage.
- Tap the action → BrandedAlert confirmation → on confirm the message is purged from the per-peer conversation cache, the NIP-04 plaintext LRU, both NIP-17 wrap caches, and the `dmInbox` state, so the row is gone from both the open thread and the Messages tab preview, and stays gone across relaunch.
- Group conversations get the same UX, backed by a new `removeGroupMessages` storage helper.

Closes #128

### Scope (per the issue body)

Local-only delete. We do **not** publish a NIP-09 deletion event yet — other parties keep their copies. A relay-side re-fetch can still re-decrypt and re-surface a kind-4 row; persistent suppression / NIP-09 publishing is tracked as a follow-up.

The action label adapts to the message direction:
- Outgoing → "Delete" + "Delete message?" / "Remove this message from your device. <peer> will keep their copy"
- Incoming → "Hide" + "Hide message?" / "The original sender keeps their copy"

### Files changed

- `src/components/MessageRow.tsx` (new) — Swipeable wrapper around `MessageBubble`.
- `src/screens/ConversationScreen.tsx` — wires `MessageRow` into the 1:1 renderItem; adds `handleDeleteMessage` that dispatches into `NostrContext`.
- `src/screens/GroupConversationScreen.tsx` — same wiring for groups, backed by the new storage helper.
- `src/contexts/NostrContext.tsx` — adds `deleteConversationMessages(otherPubkey, eventIds)` that walks all four cache layers.
- `src/services/groupMessagesStorageService.ts` — new `removeGroupMessages(groupId, ids)` helper.
- `src/services/groupMessagesStorageService.test.ts` (new) — 6 unit tests covering single + multi delete, no-op short-circuits, cross-group isolation.
- `tests/e2e/test-conversation-swipe-delete.yaml` (new) — Maestro E2E covering send → swipe → confirm → assertNotVisible.

### Why draft

Touchable behaviour deserves a real-device review before merge — gesture friction, swipe threshold, and the visual reveal of the red action panel all read differently across the AVD vs. a physical Pixel. Once the UX feels right on-device, this is ready to flip to ready-for-review.

## Test plan

- [x] `npx tsc --noEmit` — passes
- [x] `npx eslint src/ --quiet` — passes
- [x] `npx prettier --check src/` — passes
- [x] `npx jest` — 4 suites, 19 tests, all pass (6 new in `groupMessagesStorageService.test.ts`)
- [ ] Manual on-device: swipe an outgoing message, confirm Delete → row disappears, scroll the Messages tab to confirm the preview row also updates / disappears
- [ ] Manual on-device: swipe an incoming message, confirm the action is labelled Hide (not Delete) and the dialog body matches
- [ ] Manual on-device: relaunch the app and confirm the deleted row stays gone
- [ ] `maestro test tests/e2e/test-conversation-swipe-delete.yaml` on a real device or AVD (not blocking — this run requires the test fixture identities to be set up)


BenGWeeks:
Not just locally, should be deleting from the relay (although need to consider if not deleted from all relays - not sure solution to that!).

Merge conflicts.

Should also attach an image before, during (show delete button appear to the right as you swipe to the left), after deletion ("_Message delete (i)_"). The (i) could have a pop-up to explain how some relays to not respect event deletion (perhaps event give a state from your relays the message is on?). Screenshot of this too.

Related: https://github.com/BenGWeeks/lightning-piggy-mobile/pull/322

BenGWeeks:
Closing as **stale** — 187 commits behind `main`. Its swipe-`MessageRow` wrapper targeted the old inline `renderItem` in `ConversationScreen`, which was refactored into `ConversationMessageRow.tsx` (#703/#718), and `GroupConversationScreen`/`NostrContext` were split too — so the diff no longer applies. Feature still wanted (issue **#128** open); it'll be **re-implemented fresh** (swipe-to-delete/hide on `ConversationMessageRow` + the split `GroupConversationScreen` + current cache-purge points). No screenshots on this PR to preserve. Branch retained.

Issue #291, opened 2026-05-03
test(e2e): dev-mode wallet-seeder subflow to unlock wallet-conditional regression tests


BenGWeeks:
## Summary

Build `tests/e2e/_subflow-seed-wallet-as-${PIGGY}.yaml` that seeds the AVD's WalletContext with a known NWC wallet (and optionally an on-chain xpub) for the active Piggy account. Most wallet-conditional regression tests (#199, #201, #200, #144, #134, plus the entire Boltz / Send / Transfer surface) currently can't be written without this — every flow would have to walk the full AddWalletWizard inline, which is brittle and slow.

## Why filed now

The Maestro coverage audit (sub-agent run on 2026-05-03) tried to write regression flows for #199 + #201 and **couldn't, because the AVD had zero wallets configured.** Without a seeder subflow, every "happy-path" assertion fails on `extendedWaitUntil id: 'transfer-source-dropdown'`. The audit closed with 0 PRs opened because of this single missing piece.

## Approach options

- **A.** Maestro-driven AddWalletWizard walk — pure flow, no source change, but brittle + slow (~30s/seed)
- **B.** Dev-only deep link `com.lightningpiggy.app.dev://seed-wallet?nwc=...` — fast + deterministic, requires `App.tsx` linking handler + a `WalletContext.addNwcWallet` deep-link bridge
- **C.** Pre-seeded AsyncStorage snapshot via `adb shell run-as` — fastest, requires `debuggable` dev variant, brittle to schema changes

**Recommended:** Option B.

## Acceptance

- `_subflow-seed-wallet-as-{big,middle,little}.yaml` exist and run in <10s on emulator-5554
- Idempotent + composable via `runFlow:`
- Documented in `tests/e2e/README.adoc`
- Cleanup affordance: `_subflow-clear-wallets.yaml`

## Unblocks

- #199, #201, #200, #144, #134, #92, #287, #289 — every wallet-conditional regression guard

PR #322, opened 2026-05-04
feat(messages): per-message delivery indicators (WhatsApp-style ticks) (#110)

### LINKED ISSUE #110: feat(messaging): delivery indicators (WhatsApp-style ticks) on sent messages
## Summary
Show a per-message delivery state in the Conversation view so the sender can tell whether a DM / invoice card actually left the app. Today there's no visual feedback, so a message that silently fails to publish looks identical to one that landed on every relay.

## Motivation
- Seen in the wild today: user sent an invoice card that rendered in the thread, but it never appeared in Primal's DM inbox. No way to tell from the UI whether the event was rejected by relays, queued and stuck, or published but not indexed by Primal. Ticks would have caught (or narrowed) this immediately.
- Every mainstream messenger (WhatsApp, Signal, Telegram) uses tick states — users already know the vocabulary.
- Modern Nostr DM clients (0xchat, Amethyst) show per-relay send status; users now expect it.

## Proposed behaviour
Four states, rendered as small Lucide icons at the bottom-right corner of our own outgoing bubbles (DM + zap card + invoice card):

| State | Icon | Meaning |
|------|------|---------|
| Sending | `Loader2` (spinning) | `publish()` in flight, no OK from any relay yet |
| Sent | `Check` (single, grey) | ≥1 of our *write* relays returned OK for the event ID |
| Delivered | `CheckCheck` (double, grey) | Event ID observed on ≥1 of the **recipient's** read relays (from their NIP-65 list) via a follow-up `REQ` subscription |
| Failed | `AlertCircle` (red) | All publishes rejected, or no write relay returned OK within ~8s |

Tap the indicator → tooltip / bottom sheet listing the OK/failed relay URLs. Priceless when debugging "why didn't this land on Primal" scenarios.

### Stretch (v2)
- Per-recipient-relay status in the tooltip.
- `Read` state if we ever adopt a read-receipt convention (Damus has a custom kind; not widely supported).

### v1 (shippable)
- Sending / Sent / Failed — achievable purely from the existing `pool.publish()` return. No extra subscriptions.
- Icon in the outgoing card and outgoing DM bubble.
- Tooltip / long-press to see per-relay detail.

### v2 (follow-up)
- Delivered (double tick) — requires subscribing to the recipient's read relays for our event ID and caching the result.
- Accessibility: announce state via `accessibilityState` on the bubble.

## Affected files
- `src/screens/ConversationScreen.tsx` — render the tick next to each outgoing `ChatMessage`.
- `src/services/nostrService.ts` — `publishDM` / `publishEvent` should return a `{ event, relayResults: { url, status }[] }` so the UI can derive state. Currently resolves void.
- `src/contexts/NostrContext.tsx` — cache the per-event relay results so the state survives a re-render.
- New Lucide icons: `Check`, `CheckCheck`, `AlertCircle`, `Loader2` (already in `lucide-react-native`).

## Risks
- NIP-20 OK responses are optional; some relays (including LNbits' nostrclient, see #3918) don't send OK at all. Treat missing OK as "sent" not "failed" after a short timeout, otherwise we'd paint healthy sends red.
- Subscription storm: checking for every sent event on the recipient's relays adds ongoing subs. Bound the lifetime to a few minutes post-send, then give up and stay on "Sent".

## References
- WhatsApp tick semantics (single / double / blue).
- 0xchat and Amethyst tick implementations.
- Follow-up to #96 which shipped the DM + invoice card rendering.

BenGWeeks:
## Summary
- Adds a four-state delivery indicator (`sending` / `sent` / `delivered` / `failed`) on outgoing 1:1 conversation bubbles. Renders a Lucide `Check`, `CheckCheck`, or `AlertCircle` next to the timestamp.
- New state machine in `src/utils/messageDeliveryStatus.ts` is a closed transition table — `sending` is the optimistic initial state set when the bubble is appended, then resolves to `sent` or `failed` from the existing `sendDirectMessage` boolean. `delivered` is reserved for v2 (recipient-relay observation per the issue's stretch list); no v1 code path enters it. `failed` and `delivered` are terminal so a late callback from a stale promise can't regress an acked bubble.
- Tick row is suppressed on incoming messages and on historical messages loaded from the relay cache (no local-send record to attach status to). Read-receipts intentionally out of scope per Nostr's privacy model.

## Why
Today there's no UI feedback when an outgoing DM / image / invoice card silently fails to publish — it looks identical to a fully-acked send. Per the issue: "user sent an invoice card that rendered in the thread, but it never appeared in Primal's DM inbox. No way to tell from the UI whether the event was rejected by relays, queued and stuck, or published but not indexed by Primal." Ticks would have caught this.

## Files changed
- `src/utils/messageDeliveryStatus.ts` (new) — `MessageDeliveryStatus` type, `nextDeliveryStatus` guard, `deliveryStatusFromPublishResult` adapter.
- `src/utils/messageDeliveryStatus.test.ts` (new) — 11 unit tests for the transition table.
- `src/components/MessageBubble.tsx` — accept optional `deliveryStatus` prop; render a `<DeliveryTicks>` sub-component inline with the timestamp on every `fromMe` variant (text / image / invoice / lnaddr / contact / gif / location).
- `src/screens/ConversationScreen.tsx` — local `messages` state grows an optional `deliveryStatus` field; the four send paths (text, location, image, contact, GIF) now optimistically append in `sending` *before* awaiting the publish, then transition via a small `updateDeliveryStatus` helper.

## Scope
- Covers issue #110 v1 ("Sending / Sent / Failed achievable purely from the existing `pool.publish()` return"). v2 (recipient-relay observation for `delivered`) is reserved in the type but not wired.
- Group chats are unchanged for this PR — focus is the 1:1 conversation surface that the issue calls out. Group chat send is a follow-up.

## Test plan
- [x] `npx eslint src/ --quiet` clean
- [x] `npx prettier --check src/` clean
- [x] `npx jest` — all 24 tests pass (11 new in `messageDeliveryStatus.test.ts`)
- [ ] Manual: open a 1:1 chat, send a text message — bubble should appear with a single grey tick, then flip to a single white tick once the relay acks (typically <1s).
- [ ] Manual: airplane-mode, send a text message — bubble should appear with a single grey tick, then flip to a red `AlertCircle` once the publish times out, and the existing `Send failed` Alert.alert should fire.
- [ ] Manual: send a GIF, an image, a location, a shared contact, and an invoice card — all five variants should show the same indicator at the bottom-right of the bubble.

> **Draft** — UI feature, wants visual review on a real device. Did not add screenshots in this run; AVD is degraded.

Closes #110.

## Screenshot

| Bundle running this branch on emulator-5554 (Messages tab) |
|---|
| emu-pr322-bundle |

The actual delivery-indicator pixels (Check / CheckCheck / AlertCircle next to outgoing bubble timestamps) require a 1:1 conversation with a recent outgoing message visible. Triggering a fresh send hits the Amber sign flow + relay round-trip; capturing the rendered indicator deterministically needs the wallet-conditional fixture from #291. The state-machine logic is fully unit-tested in `src/utils/messageDeliveryStatus.ts`.


BenGWeeks:
Screenshot doesn't seem relevant. Can be just 1 tick to show posted, or 1 tick per relay I guess (i.e. 5 ticks if you have 5 relays).

BenGWeeks:
Closing as **stale** — 187 commits behind `main`, and its target files (`ConversationScreen.tsx` / `MessageBubble.tsx`) were refactored since (#703/#718), so the diff no longer applies. The feature is still wanted (issue **#110** open) and the delivery-status state machine here is sound — it'll be **re-implemented fresh** onto the current `ConversationMessageRow`/`MessageBubble` structure. Screenshot preserved on #110; branch retained.

PR #323, opened 2026-05-04
chore(messaging): diagnostic logs to investigate Camera/Gallery → Amber bug (#209)

### LINKED ISSUE #209: bug(messaging): tapping Camera or Gallery in AttachPanel opens Amber activity instead of attaching the image
## Symptom

In the inline AttachPanel (the WhatsApp-style icon grid that opens above the composer), tapping either:
- **Camera** tile (`attach-take-photo`)
- **Gallery** tile (`attach-send-image`)

…opens the **Amber** activity / signer screen instead of either:
- launching the device camera and capturing a photo (Camera), or
- opening the system image picker (Gallery)

…and after dismissing Amber, the image is not attached / sent.

## Expected

- **Camera** → `expo-image-picker` opens the OS camera, the captured image is then sent as a NIP-44 encrypted DM (Blossom upload then `imeta` link, per the existing `handleTakeAndSendPhoto` flow).
- **Gallery** → system image picker, then same upload + DM flow (`handlePickAndSendImage`).

## Reproduce

1. Sign in with Amber (the test fixture or any Amber-backed Nostr login).
2. Open any conversation thread.
3. Tap **+** (`conversation-attach`) to open AttachPanel.
4. Tap **Camera** OR **Gallery**.
5. Observe Amber activity opening instead of the camera / gallery picker.

## Suspected cause

The image flow probably triggers a sign event via Amber (e.g. wrapping the upload metadata, NIP-94 file event, or NIP-44 wrap of the DM payload) BEFORE the camera/gallery picker fires — so the Amber sign UI shows first and the user perceives it as the wrong destination. Or, the Amber deep-link / activity is being foregrounded synchronously when the panel handler runs (intent flag mismatch).

Worth checking:
- `handleTakeAndSendPhoto` / `handlePickAndSendImage` in `src/screens/ConversationScreen.tsx` — does anything in the call chain go through `AmberSigner` (or whatever the signer abstraction is) before `ImagePicker.launchCameraAsync` / `launchImageLibraryAsync`?
- Whether the AttachPanel's `closeAttachPanel()` + `openAttachPanel()` interactions are causing a focus/intent issue that lets Amber take over.
- Whether Amber's intent handler is registered for a scheme we're inadvertently calling (e.g. `nostrsigner://` URIs in the upload metadata pipeline).

## Reproducer environment

- Build: `8b50306` / `5bdb385` (local APK, version code 23/24 on `com.lightningpiggy.app` production)
- Device: Pixel 8 (Android)
- Signer: Amber

BenGWeeks:
## Summary

Investigation + diagnostic logging for [#209](https://github.com/BenGWeeks/lightning-piggy-mobile/issues/209). **Opened as draft** because a clean code read does not reproduce the bug — instrumentation is needed to narrow down the actual cause on the user's device.

## What I ruled out

A close trace of the tile-press → camera-launch path on both `ConversationScreen` and `GroupConversationScreen`:

```
tile.onPress → handleTakeAndSendPhoto / handlePickAndSendImage
  → setAttachPanelOpen(false)
  → ImagePicker.requestCameraPermissionsAsync() / requestMediaLibraryPermissionsAsync()
  → ImagePicker.launchCameraAsync() / launchImageLibraryAsync()
  → uploadAndSendImage(uri, base64)              <-- Amber sign happens HERE for Blossom auth (kind 24242)
    → uploadImage → uploadToBlossom → signer(...)
    → sendDirectMessage(...)                     <-- and HERE for the kind-4 / kind-1059 wrap
```

There is **no Amber invocation between the tile tap and the `ImagePicker.launch*` call**. Amber is only entered after the picker returns successfully with an asset.

I also checked:

- **`AttachPanel.tsx`** — `onPress={tile.onPress}` wires straight through, no wrapping.
- **`AmberSignerModule.kt`** — every Amber `Intent` is scoped via `intent.\`package\` = "com.greenart7c3.nostrsigner"` against `nostrsigner:` URIs. It cannot be matched against `MediaStore.ACTION_IMAGE_CAPTURE` or `Intent.ACTION_GET_CONTENT`.
- **`plugins/withAmberQueries.js`** — only adds two `<queries>` entries (`nostrsigner`, `nostr`), no broad intent filters.
- **`plugins/withNfc.js`** — explicitly comments that it does **not** register any extra intent filters on `MainActivity`.
- **`app.config.ts`** — only registers `lightningpiggy://`; no `image/*` or media intent filters.
- **PR #267** (the most recent Amber-related change) — added `sendGroupMessage` Amber support, but that fires per-recipient sign prompts only on **send**, never on tile tap.
- **`fetchConversation`** — does call `requestNip44Decrypt` (the dialog path) for fresh wraps, but that's keyed off thread mount, not off `attachPanelOpen` state.

## What this PR does

Adds two `console.log` lines on each picker entry point (4 total, gallery + camera in 1:1 + group):

- "tile pressed — requesting permission"
- "tile — launching <camera|library>"

Next time the bug reproduces, capture logcat (`adb logcat -s ReactNativeJS`) around the tile tap and we'll know whether:

1. **Both logs fire then Amber appears** → the picker launched and was occluded by a stale Amber intent (look at the Amber pending-promise state in `AmberSignerModule.kt`).
2. **Only the first log fires** → something between permission and launch is racing in front of the picker.
3. **Neither fires** → the tap is being intercepted before reaching this handler — most likely the Amber UI was already up from an unrelated background sign request and the user's tap landed on Amber's surface.

## Test plan

- [ ] Land this, build, reproduce on the user's Pixel 8 with Amber.
- [ ] Capture `adb logcat -s ReactNativeJS:V *:S` while tapping Camera then Gallery in a thread.
- [ ] Decide on the real fix from the log evidence and follow up with a targeted PR.

## Why draft

No functional fix yet — only telemetry. Ready-for-review once the logs from a reproduction confirm where Amber actually enters the flow.

Closes #209


BenGWeeks:
Closing as **obsolete** (superseded by refactor). This draft only added 4 `console.log` diagnostic lines to investigate the Camera/Gallery-opens-Amber bug (#209) — no functional fix. The handlers it patched (inline `handlePickAndSendImage`/`handleTakeAndSendPhoto` in `ConversationScreen`/`GroupConversationScreen`) were since **extracted and unified into `src/hooks/useComposerActions.ts`** (#703/#718, then reworked by #299/#688/#729), the branch is 187 commits behind / conflicting, and none of its logging ever landed. Issue **#209 stays open** — if we want logcat evidence again, instrument the single shared hook on demand.

Issue #31, opened 2026-04-03
Allow creating a new Nostr identity (npub/nsec)


BenGWeeks:
## Description

Users who don't have a Nostr identity should be able to create one directly in the app instead of needing to use an external tool.

## Acceptance Criteria

- [ ] "Create New Identity" option on the login sheet (alongside "Paste nsec" and "Use Amber")
- [ ] Generate new keypair using nostr-tools
- [ ] Prompt user to set display name and optional profile picture
- [ ] Publish kind 0 (metadata) event to default relays
- [ ] Store nsec securely in expo-secure-store
- [ ] Show the user their npub and nsec with a strong warning to back up the nsec
- [ ] Option to copy/export nsec for backup
- [ ] Maestro E2E test for the full create-account flow

BenGWeeks:
Related: #5 (Nostr login - prerequisite infrastructure), #30 (edit profile - natural next step after identity creation). Being addressed in draft PR #22.

PR #335, opened 2026-05-04
perf(nostr): persistent relay subscription for DMs alongside refreshDmInbox (#188)

### LINKED ISSUE #188: perf(nostr): replace one-shot querySync with persistent relay subscription for DMs
## Problem

Every `refreshDmInbox` call and `fetchConversation` call opens a fresh `pool.querySync` round-trip (`nostr-tools/pool`). On a relay-pool of 4 endpoints each 15 s-capped, that's a minimum 1-3 s of wire latency per inbox refresh / thread open, even on fast wi-fi. On 4G cellular it's 3-8 s.

This means:
- Messages only appear when the user pulls-to-refresh or switches tabs (whichever retriggers `useFocusEffect`).
- Every relay query re-fetches the same events from the same relays, throttled only by the new `since:` filter (PR #???).
- Real-time DM feel is impossible — a message sent while the user is staring at the Messages tab only appears on the next TTL-expired focus.

## Fix

Replace the `querySync` pattern with a persistent `pool.subscribeMany` that stays open for the lifetime of the tab / screen. Outline:

1. Open a subscription on app launch (or Messages-tab mount) with the same filter shape currently used for `fetchInboxDmEvents`.
2. Accumulate events into an in-memory `Map<id, event>` as they stream in.
3. On thread open, read from that map (instant) and open a narrower sub filtered to `#p:[peer]` that overrides until the thread closes.
4. Close subs on tab blur / logout to free relay sockets.

Reference: Damus uses a single `advancedStream` at launch (`damus/Nostr/SubscriptionManager.swift`), Coracle uses per-channel `AbortController`-scoped subs (`engine/requests.ts:259-287`). Pattern details in the research summary attached to the perf PR.

## Caveats

- `@getalby/sdk` NWC subscription is orthogonal (same pool but different filter set) — reuse the same `SimplePool` instance.
- Clock drift pad (`since - 120`) from the current `fetchInboxDmEvents` stays relevant for re-subscribe after network hiccups.
- Parental-control follow filter at `NostrContext.tsx:1445` must still apply; just move it to the subscription-event handler.

## Estimate

~4 h — architectural change. Touches `nostrService.ts` + `NostrContext.tsx`. Non-trivial but no API breakage for consumers.

## Related

Builds on the perf PR landing now (LRU cache + `since` delta + persisted inbox/conv). This is the next step — the current PR shortens payloads; this one removes the polling entirely.

BenGWeeks:
## Summary

- Adds `subscribeInboxDmEvents` (`nostrService`) — a long-lived `pool.subscribeMany` over kind-4 (sent + received) and kind-1059 (NIP-17 wraps p-tagged to the viewer). Mirrors the filter shape of `fetchInboxDmEvents` so it can be processed through the same path.
- `NostrContext` starts the sub on hydrate / login, tears it down on logout / identity change / unmount. On each incoming event it fires a debounced (500 ms) force-refresh through `refreshDmInbox` so streamed events get the same NIP-17 unwrap + cache-merge treatment the cold-start path uses.
- `querySync` cold-start path is **untouched** — this is option B (additive), per the issue body's risk note.

## Why option B?

A full migration (rip out `querySync`) would require duplicating the entire NIP-17 unwrap + group-routing + signer-specific decrypt + LRU cache pipeline inside the sub callback, and subscription bugs are notoriously hard to spot in code review. The additive path delivers the issue's UX win — real-time messages without waiting for tab focus / pull-to-refresh — at near-zero behavior-loss risk. A future PR can phase out `querySync` once the persistent sub has soaked.

## Lifecycle

- **Start**: `useEffect` keyed on `[isLoggedIn, pubkey, getReadRelays, refreshDmInbox]`. Fires once the user is hydrated and read relays are known.
- **Stop**: cleanup function closes every underlying sub (`subSentK4`, `subRecvK4`, `subWraps`) and clears the debounce timer. Triggered by logout, identity change, unmount, or relay-set change.
- **Reconnect**: delegated to `SimplePool` — its relay layer re-emits past events on socket reconnect, so we don't manage WS lifecycle ourselves. A relay-set change closes the old sub set before opening a new one against the new relays.
- **No quota leak**: every `subscribeMany` call is matched by an `s.close()` on teardown — covered by a unit test.

## Dedupe story

The persistent sub firing the same wrap that `refreshDmInbox` is also processing is handled by **two existing mechanisms**:

1. **Wrap-id NIP-17 cache** (`AMBER_NIP17_CACHE_KEY` / `NSEC_NIP17_CACHE_KEY`) — the second processing of a wrap hits the cache (no decrypt, no Amber IPC), making the cost negligible.
2. **`mergeInboxEntries` id-keyed merge** — collapses duplicate inbox entries on event id. The same wrap landing through both the sub trigger and a concurrent `fetchInboxDmEvents` collapses to one inbox row.

`dmInboxInFlight` (single-flight) further coalesces refresh triggers from event bursts so a 3-message burst becomes one refresh, not three.

## Test plan

- [x] `npx eslint src/ --quiet` — clean
- [x] `npx prettier --check src/` — clean
- [x] `npx jest` — 19/19 passing (6 new tests for the sub: filter shape, kind-routing, teardown closes all subs, throw-resilient unsubscribe, re-subscribe opens fresh subs, handler-throw isolation)
- [ ] Real-device: messages from a peer arrive in inbox without tab focus / pull-to-refresh (verify via grep `[Nostr] starting persistent DM subscription` in logcat)
- [ ] Real-device: log out → log back in → confirm one `starting…` and one `stopped…` line per session (no leaked sub from previous identity)
- [ ] Real-device: toggle a relay off in Settings → confirm sub re-fires against the new relay set
- [ ] Real-device: app background / foreground → relay sockets resume and events still stream (SimplePool reconnect)

Closes #188

### DIFF
--- a/src/contexts/NostrContext.tsx
+++ b/src/contexts/NostrContext.tsx
@@ -2416,6 +2416,89 @@ export const NostrProvider: React.FC<{ children: React.ReactNode }> = ({ childre
     if (!isLoggedIn) setDmInbox([]);
   }, [isLoggedIn]);
 
+  /**
+   * Persistent relay subscription for inbound DM events (#188).
+   *
+   * Layered ALONGSIDE `refreshDmInbox` per the issue's option B: the
+   * one-shot `querySync` still drains historical state on cold start /
+   * pull-to-refresh; this subscription catches anything published
+   * after that, eliminating the focus-driven polling loop.
+   *
+   * Design notes:
+   *  - Lifecycle: starts when the user is hydrated (`pubkey` + `isLoggedIn`)
+   *    and read relays are known; tears down on logout / identity change /
+   *    unmount. The teardown closes every relay sub it opened so we don't
+   *    leak relay quota across logins.
+   *  - Reconnect: handled inside `nostr-tools/SimplePool` — the subs are
+   *    long-lived and the pool re-emits past events on socket reconnect.
+   *    A relay set change (read-relays toggle, NIP-65 update) re-fires
+   *    this effect, which closes the old sub before opening the new one.
+   *  - Dedupe: every incoming wrap fires a debounced
+   *    `refreshDmInbox({ force: true })`. The wrap-id NIP-17 cache (the
+   *    same one the cold-start path uses) makes re-processing cheap — a
+   *    cache hit per wrap, no double decrypt. `mergeInboxEntries` dedupes
+   *    on event id, so the same wrap landing through both the sub and a
+   *    concurrent `fetchInboxDmEvents` call collapses to one inbox row.
+   *  - Single-flight: `refreshDmInbox` already coalesces overlapping
+   *    refreshes via `dmInboxInFlight`; bursts of arrivals through the
+   *    sub piggy-back on the in-flight refresh instead of spawning
+   *    duplicate work.
+   *  - Debounce: 500 ms — a typical message burst (e.g. a friend
+   *    splitting a paragraph into 3 wraps) collapses into a single
+   *    refresh tick instead of 3 back-to-back ones.
+   */
+  const dmSubDebounceTimer = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const PERSISTENT_DM_SUB_DEBOUNCE_MS = 500;
+  useEffect(() => {
+    if (!isLoggedIn || !pubkey) return;
+    const readRelays = getReadRelays();
+    if (readRelays.length === 0) return;
+
+    const scheduleRefresh = (): void => {
+      if (dmSubDebounceTimer.current) clearTimeout(dmSubDebounceTimer.current);
+      dmSubDebounceTimer.current = setTimeout(() => {
+        dmSubDebounceTimer.current = null;
+        // Force-refresh so the TTL gate doesn't swallow the streamed
+        // event. The single-flight guard means a refresh already in
+        // progress will absorb this trigger instead of racing.
+        refreshDmInbox({ force: true }).catch((err) => {
+          if (__DEV__) console.warn('[Nostr] persistent-sub refresh failed:', err);
+        });
+      }, PERSISTENT_DM_SUB_DEBOUNCE_MS);
+    };
+
+    if (__DEV__) {
+      console.log(
+        `[Nostr] starting persistent DM subscription on ${readRelays.length} relay(s) for ${pubkey.slice(0, 8)}…`,
+      );
+    }
+    const unsubscribe = nostrService.subscribeInboxDmEvents({
+      myPubkey: pubkey,
+      relays: readRelays,
+      onKind4: (_ev) => scheduleRefresh(),
+      onKind1059: (_ev) => scheduleRefresh(),
+    });
+
+    return () => {
+      if (dmSubDebounceTimer.current) {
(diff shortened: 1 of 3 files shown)


BenGWeeks:
Skipping rebase — substantial structural conflict needs human review.

NostrContext.tsx has gained ~480 lines of new logic on main since this branch (live-sub-since-cursor in #408, NIP-46 inbox decrypt in #284's branch path, kind-4 cursor pinning in #408's follow-up). The persistent-DM-subscription approach this PR proposes overlaps directly with that work — significant rethink needed to layer them coherently rather than re-applying the original intent on top.

Recommend either:
1. **Reset and re-implement** — squash the original idea, drop the existing diff, re-apply against current main with awareness of what already landed.
2. **Defer** — main's recent live-sub work (#408, #412) may have already absorbed enough of this PR's value to make it redundant. Worth checking whether the persistent-subscription concern is still felt before re-investing.

Either's fine; my read is option 2 is likely correct (the live-sub-since-cursor work is exactly the kind of persistent subscription this PR was scoped at) but you'd know for sure.

BenGWeeks:
## Superseded by work already on main

After landing the perf stack for #31, I went to merge this PR and discovered main has independently implemented a more refined version of the same idea. Specifically:

- **#508** — `perf(nostr): carry knownWrapIds across live-sub re-opens + eager add` (commit `16aee90`)
- **#510** — `perf(boot): cut cold-start tap latency 12s → ~1s on real Pixel` (commit `1d5fe8a`)

Main's inline live-DM-sub implementation in `NostrContext.tsx` (lines ~3500-4150, ~650 lines) covers everything this PR intended *plus*:

- **Lazy `liveSubArmed` gate** — sub only opens after the user actually focuses a DM surface (Messages tab / ConversationScreen). Home-only users on cold boot don't pay any per-wrap unwrap/route cost. This PR's option B always opened the sub on hydrate, which is the issue #510 was solving.
- **Wrap-id dedup persisted across sub re-opens** (`knownWrapIdsRef`) — when the relay-list refresh fires the effect again 9 s into cold start, the re-streamed backlog of 100+ wraps short-circuits at the in-memory Set instead of re-decrypting / re-routing each one. This was issue #505.
- **Inline NIP-17 unwrap + group-routing** — kind-1059 wraps decrypt and route to group / 1:1 in the sub handler itself, instead of triggering a debounced `refreshDmInbox`. Removes the double-process round trip this PR was relying on.
- **Batched `setDmInbox` flushing** — 25 entries or 150 ms per batch, so a cold-start drain doesn't fire 30 setStates/sec.
- **Follow gate** — mirrors `refreshDmInbox`'s B1 invariant inline.
- **`subscribeInboxDmEvents` helper** — never reached main; main wires `pool.subscribeMany` directly in the useEffect.

The 6 unit tests in this PR cover `subscribeInboxDmEvents` (the standalone helper). With the helper unused on main, the tests would have no production code to guard.

### Issue #188 is still open

Main's #510 + #508 stack does resolve #188 in practice (persistent sub replaces the focus-driven polling), but #188 wasn't marked closed. Worth a separate housekeeping pass to close #188 with links to #510 + #508.

Closing this PR.

PR #365, opened 2026-05-04
fix(groups): show stacked member avatars on Groups list (#363)

### LINKED ISSUE #363: Groups with 2 members should show 2 profile icons (a group of 3 is OK)
## What

In the Groups list (Friends → Groups), every entry currently renders a **single** circular avatar with the group name's first letter (e.g. `T` for `TestGroup-15:08:26`), regardless of how many members the group has.

For groups with **2 members**, this isn't very informative — you can't tell who's in it without opening the thread. The expected pattern (and a more useful one) is to show **two stacked / overlapping profile pictures**, one per member, sourced from each member's kind-0 avatar.

For groups of **3 or more members**, a single composite/initial icon is fine — there isn't enough room to render N avatars cleanly.

## Repro

1. Sign in as Big Piggy.
2. Open Friends → Groups.
3. Note any 2-member group: the icon is a generic letter circle, not the two members' avatars.

## Expected

- 2-member group: two overlapping circular avatars showing each member's profile picture (or initial fallback if no kind-0 picture).
- 3+ member group: existing single composite/initial icon is fine.

## Notes

- Same pattern is already used elsewhere in the app for friend pickers / picker chips — could likely reuse the avatar-pair component.
- Falls back gracefully when one or both members have no kind-0 picture: render initial-circle for the missing one(s).

## Screenshot

See attached image — the two circled entries are the 2-member groups that should show the two member avatars instead of the `T` letter circle.

BenGWeeks:
## Summary

Closes #363.

The Groups list (Friends → Groups) rendered a single brand-pink letter circle per row regardless of member count, so 2-member and 3-member groups looked identical and you couldn't tell who was in them without opening the thread.

This PR:

1. **Replaces the letter-circle avatar with the existing `GroupAvatar` primitive** (already used by `GroupRow.tsx` in the Messages tab), so each row shows up to 3 stacked member profile pictures.
2. **Prepends the viewer's pubkey to the avatar cluster** in both `GroupsScreen` and `GroupRow`, so the cluster size matches the actual people-count in the group (matters when no one has posted yet — without the viewer prefix, an inactive 1:1 group would render a single avatar and a 3-person group only 2).
3. **Aligns GroupsScreen's "{N} members" subtitle to the `+1` convention `GroupRow` already uses** so the same group reads with the same count on both screens, and the displayed count matches the avatar count.

`memberPubkeys` excludes the viewer by LP convention (`groups context line 345`), so these three changes bring the visual consistently in line with how the rest of the app talks about group sizes.

## Before / after

| | Before (letter circle, count off vs. Messages tab) | After (stacked avatars + matching count) |
|---|---|---|
| Groups list | before-groups-list | after-groups-list |
| Messages tab (same fix in `GroupRow`) | – | after-messages-tab |

`GroupAvatar`'s existing slot-position rules now drive the visual:

- `pubkeys.length === 1` → 1 centered avatar (e.g. 1:1 chat where only the other member resolves to a kind-0)
- `pubkeys.length === 2` → 2 diagonally-stacked avatars (the case the issue specifically called out)
- `pubkeys.length >= 3` → 3 stacked avatars (top-left, top-right, bottom-centre); only the first 3 are rendered

When a member doesn't have a kind-0 picture cached, `GroupAvatar` falls back to a `UserRound` placeholder per slot — visible in the "after" screenshots for the test-fixture members that don't yet publish kind-0 images.

## Implementation notes

- `GroupsScreen.tsx`: build the `contactInfoMap` once per render (mirrors `MessagesScreen`'s O(contacts) idiom #245), pass `[myPubkey, ...item.memberPubkeys]` plus the map into `GroupAvatar`. Drop the now-unused `.avatar` / `.avatarLetter` styles. Bump the count subtitle to `memberPubkeys.length + 1`.
- `GroupRow.tsx`: change the second loop's source from `group.memberPubkeys` to `[myPubkey, ...group.memberPubkeys]` so the cluster fills correctly when there are no recent senders.

## Test plan

- [x] `npx eslint` clean for the two modified files
- [x] `npx prettier --check` clean
- [x] Manual on emulator (LP_API34, x86_64 API 34) via Metro:
  - 3-person groups (2 in `memberPubkeys`) render 3 stacked avatars and read "3 members" — matches `GroupRow` on Messages tab
  - 1:1 groups (1 in `memberPubkeys`) render 2 stacked avatars (viewer + other) and read "2 members"
  - 4+ person groups cap at 3 avatars with the bottom-centre slot
- [ ] Manual on Pixel via dev build with real Nostr identities (mostly to confirm kind-0 pictures actually resolve)

## Notes

This PR doesn't touch the Piggy test-fixture kind-0 profile images themselves — that's a separate concern (some `Piggy` identities don't publish a `picture` field yet so the rows show the `UserRound` placeholder instead of a tinted piggy avatar). Will address in a follow-up.


BenGWeeks:
Addressed both Copilot comments in commit $(git rev-parse HEAD | cut -c1-7) — derived both the displayed count and `GroupAvatar` pubkeys from one precomputed `groupAvatarInputs: Map<groupId, string[]>`, so:

- The count and cluster size can never disagree (incl. when `myPubkey` is null during session restore).
- The pubkey arrays are stable across unrelated re-renders so `React.memo(GroupAvatar)` actually skips work.

Threads resolved.

PR #533, opened 2026-05-13
perf(dm): time-budget yields + hard-cancel decrypt queue (#532)

### LINKED ISSUE #532: perf(dm): NIP-17 decrypt loop starves JS thread; tab-switch + first-login feel slow
## Symptoms

Reported on PR #488 review:

> if i pull to refresh on messages, then navigate to explore, it becomes very slow. Also takes a long time to load messages on the first load after logging in.

Two repro paths, same root cause:

**A. Refresh → tab-switch**
1. On Messages tab, pull-to-refresh.
2. Before the refresh completes, tap the Explore tab.
3. Explore home renders sluggishly — animations stutter, rails appear in slow motion.

**B. First-login**
1. Fresh sign-in (Amber or nsec).
2. Splash sits for 10-30+ seconds with no visible progress.
3. Tab bar is unresponsive during the freeze.

## Root cause

`refreshDmInbox` in `src/contexts/NostrContext.tsx` decrypts every NIP-17 kind-1059 wrap on the JS thread. `DECRYPT_YIELD_EVERY` yields *between batches* but each individual libsodium decrypt + JSON.parse + rumor-build runs synchronously on-thread, so the UI thread is grabbed for ~5-20 ms per wrap.

`MessagesScreen` already aborts the in-flight refresh on blur (see #412 / the comment at MessagesScreen.tsx:173), but the AbortController only stops *the outer loop*. Work that's already been enqueued via `setImmediate` continues to drain after blur — usually for several seconds — so the next screen's mount lands on a half-blocked thread.

On first login this is amplified ~100×: the user has hundreds of historical wraps to decrypt before the inbox is interactive, with no parallelism.

## Possible fixes (ranked by impact / cost)

1. **Move the decrypt loop off the JS thread.** Either expo-modules-core's `runOnNewThread`, a dedicated JSI worker, or react-native-worklets-core. Each wrap goes 5-20 ms → ~0 ms on the UI thread. Highest impact, highest effort.
2. **Stronger cooperative yielding.** Yield every wrap (not every N), capped at e.g. 4 ms decrypt budget per frame. Single-line change, immediate UX improvement, but doesn't help wall-clock throughput on first login.
3. **Hard-cancel the work queue on blur.** Drain the pending setImmediate queue when the AbortController fires, not just stop scheduling new chunks. Fixes the \"tab-switch feels slow\" symptom even if first-login is still slow.
4. **Persist decrypted rumors** so a second login (or app restart) doesn't redo the work. Already partially in place (rumors → AsyncStorage); audit whether the cache is being hit consistently on cold start.
5. **Background-prefetch DM wraps before login completes** so by the time the user is interacting with Home, the inbox is already partially populated.

## Suggested PR scope

Start with (3) + (2) for a quick low-risk win, then evaluate whether (1) is needed. (1) unlocks real parallelism but is enough native-bridge churn that it's worth landing on its own.

## Related

- #531 — Login progress dialog (UX-side fix for the freeze, complementary)
- PR #488 — Explore tab (where the symptom was reported)
- Issue #412 — Earlier work on aborting refreshDmInbox on tab blur (partial fix)

BenGWeeks:
## Summary

Targeted at issue #532 — the NIP-17 inbox decrypt loop holds the JS thread in 50–200 ms bursts, making tab-switch + first-login feel sluggish. This PR lands the two low-risk fixes from the issue's suggested scope (items 2 + 3); the worker-thread / parallelism work (item 1) is deferred to a follow-up.

Changes are confined to `src/contexts/NostrContext.tsx`:

- **Time-budget yields.** New `createYieldScheduler(...)` yields whenever wall-clock since the last yield exceeds **`DECRYPT_FRAME_BUDGET_MS = 4 ms`** (half a 60 fps frame). The existing `NIP17_LOOP_YIELD_EVERY = 4` is retained as a safety cap. A run of cheap cache hits no longer pays for a `setTimeout(0)` every 4th iteration when there's been no blocking work.
- **Hard-cancel on abort.** The scheduler attaches an `abort` listener that `clearTimeout`s the in-flight pending handle and rejects the pending promise, so the loop unwinds in the next microtask instead of waiting out one more scheduler tick. Fixes the "Explore tab feels slow right after Messages pull-to-refresh" symptom.
- Both the nsec and Amber NIP-17 branches now make a single `scheduler.maybeYield()` call per iteration in place of the two prior modulo-based yield sites.

## Out of scope (deliberately)

- Worker thread / `runOnNewThread` decrypt (issue item 1) — bigger change, lands separately.
- Cache persistence audit (item 4) and background prefetch (item 5).
- The kind-4 batch loop + `fetchConversation`'s decrypt batch — both already gate on `Promise.all` per chunk with abort checks between.
- `MessagesScreen.tsx` blur-abort hook is correct upstream of the loop; not touched.

## Observability

`[Perf] nip17-cache` log now includes `yields=N (budget=4ms, cap=4)` so the new behaviour is observable post-merge.

## Test plan

- [x] `npx prettier --check src/contexts/NostrContext.tsx` clean
- [x] `npx eslint src/contexts/NostrContext.tsx` — only the pre-existing `Array<T>` warning at line 1638 (not touched by this PR)
- [ ] Maestro perf measurement (`scripts/perf-*.sh`) on AVD + Pixel: legacy-jank + p99 frame time, before/after, refresh-then-tab-switch flow. **I'll add the numbers from the main checkout — agent isn't running Maestro per the no-sub-agents-for-emulator rule.**
- [ ] Manual: pull-to-refresh on Messages, immediately tap Explore — confirm Explore rails animate smoothly, no stutter.

Closes #532

### DIFF
--- a/src/contexts/NostrContext.tsx
+++ b/src/contexts/NostrContext.tsx
@@ -335,6 +335,122 @@ function yieldToEventLoop(): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, 0));
 }
 
+/** Half of a 60 fps frame (~8.3 ms). The NIP-17 inbox loops aim to
+ * stay under this many ms of unbroken JS work per yield. With the
+ * old count-based yield (every 4 wraps) a slow path could still
+ * blow past 50–200 ms in a single burst — enough to drop several
+ * frames on tab-switch. See #532. */
+const DECRYPT_FRAME_BUDGET_MS = 4;
+
+/** Cooperative-yield scheduler for the NIP-17 inbox loops (#532).
+ *
+ * Two improvements over the old `if (i % N === 0) await yieldToEventLoop()`
+ * pattern:
+ *
+ * 1. **Time-budget yields.** We only pay for a `setTimeout(0)` round-
+ *    trip when wall-clock since the last yield exceeds
+ *    `DECRYPT_FRAME_BUDGET_MS`. A run of cheap cache hits no longer
+ *    forces a yield every Nth iteration even though there's been no
+ *    blocking work. The count-based modulo still acts as a safety
+ *    cap (set by the caller) so a pathological iteration that
+ *    somehow underestimates its own runtime can't starve the thread.
+ *
+ * 2. **Hard-cancel on abort.** When the caller's `AbortSignal` fires
+ *    mid-loop, an `abort` listener clears the currently-pending
+ *    `setTimeout` and resolves the awaiter immediately. Without this,
+ *    the loop would still drain one more `setTimeout(0)` round-trip
+ *    (plus whatever sync work follows it) before the next abort
+ *    check — visible as a slug of pinned-thread time after a
+ *    tab-switch blurs MessagesScreen.
+ *
+ * Returned object:
+ * - `maybeYield()` — call once per loop iteration. No-op unless the
+ *   frame budget is exceeded OR the safety-cap counter ticks.
+ * - `yieldCount` — number of actual yields performed (perfLog).
+ * - `dispose()` — detach the abort listener after the loop exits.
+ */
+type YieldScheduler = {
+  maybeYield: () => Promise<void>;
+  readonly yieldCount: number;
+  dispose: () => void;
+};
+
+function createYieldScheduler(opts: {
+  signal?: AbortSignal;
+  /** Safety cap — always yield when iteration % safetyEvery === 0,
+   * even if the frame budget hasn't been blown. */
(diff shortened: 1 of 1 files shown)


BenGWeeks:
@copilot — please review. This is the cheap-win pass on #532: time-budget yielding in the NIP-17 decrypt loop + hard-cancel of the pending setTimeout on abort. Single-file change in NostrContext.tsx. Worker-thread parallelism is queued as a follow-up if these don't move the needle enough.

Issue #561, opened 2026-05-15
perf(wallet): migrate to NIP-47 notifications instead of polling for balance/payments


BenGWeeks:
Tonight's [#560](https://github.com/BenGWeeks/lightning-piggy-mobile/issues/560) debugging showed the dominant boot/runtime cost is the NWC `getBalance` polling:

- 3+ parallel `getBalance` calls fire at app boot (initial connect + refreshActiveBalance + various effect-driven refreshes), each retrying 2× on the flaky 'no info event (kind 13194) returned from relay' error → ~6-18 s of wakefulness per boot.
- The `expectPayment` 1 s poll fires `lookupInvoice + getBalance` every tick for the full 3 min wait window after a claim, with each `getBalance` blocking 1-3.6 s when the relay is slow → silent 20-45 s JS-thread freezes from the resulting render storm.

Ben's suggestion (right call): **NIP-47 notifications**. The wallet service pushes `payment_received` and `payment_sent` events directly to subscribed clients. We listen once and react to events as they arrive — no polling, no retry storms.

## Plan

- Confirm the user's NWC server (LNbits) advertises `notifications` in its `get_info` response capability list.
- Subscribe to NIP-47 notification events on wallet connect via @getalby/sdk's notification API (kind 23196 / 23197 wallet → app events).
- Replace the 1 s `expectPayment` tick with a notification listener gated on `paymentHash`. Keep a short fallback poll (~10-30 s cadence) for wallets that don't support notifications.
- Replace the 30 s active-wallet `getBalance` poll with a `balance_updated` subscriber (if the spec / server supports it) or compute from `payment_received` / `payment_sent` deltas.
- Keep `getBalance` for explicit user-initiated refresh + initial-connect parity check.

## Acceptance

- After connecting an NWC wallet on a clean boot, fewer than 2 `getBalance` requests fire in the first 30 s.
- Claiming a Piggy doesn't spawn a 1 s poll. `payment_received` notification arrival is what fires the success toast.
- The dev-client `[Perf]` heartbeat shows no gap > 1 s during the 3-minute post-claim window on the Pixel.

Today's dedupe stop-gap landed in nwcService — this issue is the proper architectural follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #560, opened 2026-05-15
perf(hunt): 30s JS-thread lockup on HuntPiggyDetail / Explore cold mount


BenGWeeks:
Recurring issue: navigating to either the Explore tab or a HuntPiggyDetail screen pins the JS thread for 15–30 s on the Pixel dev client. Symptoms:

- Heartbeats stop firing for 20–30 s at a time
- Cache detail loading spinner persists; `Scan the Piglet` taps queue but don't process
- Often co-located with NWC reconnect bursts (now de-duped via commit 2c9579e but not eliminated)
- Smells of compounding causes: subscribeNearbyCaches event burst + contact-list hydration + profile fetches + WebView mount overlap

Already addressed (partial):
- a6193cd: nearby-rail PTR no longer wipes caches Map
- 4814e8d: fetchCachesByAuthor 5 s maxWait
- 2c9579e: NWC reconnect dedupe
- 7b3b041: HuntPiggyDetail seeds from local cache mirror before relay fetch

Still seen on the Pixel after all four fixes — needs a proper perf pass with the existing scripts/perf-*.sh harness on AVD + Pixel to capture the freeze window and find the actual hot path.

Sibling of #30. Closes when navigating from cold Home → Explore → HuntPiggyDetail and back takes p95 < 3 s on a clean dev-client install on the Pixel.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

BenGWeeks:
Update from tonight's Pixel session (Ben on dev client, end-to-end NFC claim flow working):

Captured freeze gaps from `[Perf] heartbeat` log entries:
- gap=23990ms between heartbeats #24 and #25
- gap=45153ms between heartbeats #25 and #26
- gap=2399ms between heartbeats #26 and #34

Critically: **no ReactNativeJS log lines fired in those windows.** Whatever is blocking the thread runs in code without `[PerfBlock]` instrumentation, so logs alone can't pinpoint it.

Symptoms preceding the freeze:
- `[NWC] Connection lost, reconnecting...` (despite the 2c9579e dedupe fix — connection itself is unstable, dedupe just prevents parallel reconnects)
- `'Failed to request get_balance', [Error: no info event (kind 13194) returned from relay]`
- `[NWC] getBalance(...) attempt 1 failed, retrying in 1500ms...`
- Then 24s+45s of dead air

Hypotheses worth instrumenting next session:
[...]

PR #562, opened 2026-05-16
perf: React.Profiler + tighter NIP-17 yield to surface 24-57s freeze (#560)

BenGWeeks:
Closes part of #560.

## What

Two complementary changes targeting the 20-57 s JS-thread freezes Ben hit on the Pixel:

1. **`NIP17_LOOP_YIELD_EVERY: 4 → 2`** — `refreshDmInbox` was running 8.6 s wall-clock with 3 s heartbeat gaps stacking during the unwrap loop. Halving the yield cadence cuts per-burst blocking back to ~2 ms; the setImmediate cost amortises across the still-significant per-wrap decrypt work so overhead is < 5%.

2. **React.Profiler wrappers around HomeScreen + ExploreHomeScreen** — tonight's logs showed the dominant freeze is in code without `[PerfBlock]` markers. The most likely culprit is render-commit cost from the 596-contact `setContacts` dispatch triggering tree-wide re-renders. Profiler measures the commit phase directly; `onRender` threshold of 100 ms (user-perceived jank floor) keeps logcat quiet on healthy frames and only fires on real outliers.

Output as `[PerfBlock] render:<id> <phase>=<ms>ms`, greppable alongside the existing PerfBlock markers (`subscribeNearbyCaches`, `NWC.getBalance`, etc.).

## Why this PR exists separately from #488

#488 already carries a lot of feature work for the Explore tab; the perf instrumentation here is targeted at the residual freeze that's not directly tied to the new surfaces. Keeping it on a sibling branch makes the diff readable and the perf-vs-feature decision easier to revert independently.

## Test plan

- [ ] Re-run Pixel perf suite after merge + compare heartbeat gap distribution
- [ ] On a Pixel session: navigate Home → Explore → cache detail → back, and verify each render takes < 100 ms or shows up in logcat with its cost

## Follow-ups (NOT in this PR)

- The proper fix is GH #561 (NIP-47 notifications instead of polling) — once the wallet service pushes `payment_received` / `balance_updated` events, the polling that drives most of the NWC retry storm goes away.
- If Profiler shows render-storm is the dominant cost, the next step is memo-ising heavy children of HomeScreen / ExploreHomeScreen (TransactionList, WalletCarousel, etc.).

### DIFF
--- a/src/contexts/NostrContext.tsx
+++ b/src/contexts/NostrContext.tsx
@@ -484,8 +484,15 @@ const DECRYPT_YIELD_EVERY = 15;
  * breathing room for gorhom-bottom-sheet's open animation to schedule
  * frames. Halving this doubles yield frequency, drops the per-burst
  * blocking from ~8 ms to ~4 ms, and lets bottom-sheet opens stay
- * smooth during inbox drain. */
-const NIP17_LOOP_YIELD_EVERY = 4;
+ * smooth during inbox drain.
+ *
+ * Lowered again 2026-05-16 from 4 → 2: tonight's instrumented Pixel
+ * logs (issue #560) showed refreshDmInbox running for 8.6 s wall-clock
+ * with 3 s heartbeat gaps stacking during the decrypt loop. Yielding
+ * every 2 wraps cuts each per-burst block back to ~2 ms; the
+ * setImmediate cost is amortised across the still-significant
+ * per-wrap decrypt work so the overhead is < 5%. */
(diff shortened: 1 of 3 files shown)


BenGWeeks:
Copilot's two Profiler-gating comments are valid. Keeping Profiler ungated for now so it remains active in preview builds where we're actively chasing the JS-thread freezes (#560). Tracked under #564 as part of a broader pre-release logging cleanup pass before the next production cut.

Issue #552, opened 2026-05-15
feat(explore): migrate Leaflet-in-WebView to MapLibre Native (vector tiles, OSM, no Google)


BenGWeeks:
## Background

All three map surfaces — `MapScreen`, `ExploreMiniMap`, `LocationPickerSheet` — currently render Leaflet inside a `react-native-webview`. The pattern has served well but has accumulated real friction:

- **Bridge round-trip on every data update.** Pins live in the WebView; React state changes have to be `injectJavaScript`-ed across the bridge. Cold-start specifically: even when merchants are seeded synchronously from disk cache (#15 / PR #550), the pins still wait on the Leaflet CDN fetch → `ready` postMessage → React `useEffect` round-trip before they appear. Fixed in `fca4b01` by baking the initial payload into the HTML, but the underlying impedance mismatch remains.
- **Gesture quirks.** Inline maps inside scrollable parents needed an `onInteractionChange` callback to stop the parent ScrollView from stealing vertical drags (#26 / committed). Native maps wouldn't have this class of bug.
- **No first-class marker components.** Pin / circle / popup definitions live as JS strings inside `makeHtml(…)`, with backtick-in-comment traps and template-literal escaping.
- **CDN dependency.** Leaflet currently loads from `unpkg.com` on every cold start.

## Proposal

Migrate to **[MapLibre Native](https://maplibre.org/)** via `@maplibre/maplibre-react-native`.

### Why MapLibre specifically

| Option | Tile source | API key needed | OSM-aligned? | Marker model |
|---|---|---|---|---|
| **MapLibre Native** | Stadia / Protomaps / MapTiler free / self-host | None for free tiers | ✅ | First-class React components |
| react-native-maps | Google Maps (Android) / Apple Maps (iOS) | **Google API key required** | ❌ | First-class React components |
| Mapbox GL (`@rnmapbox/maps`) | Mapbox (commercial) | Mapbox token | Partial (commercial OSM derivative) | First-class React components |
| Leaflet-in-WebView (status quo) | OSM raster | None | ✅ | JS string injection |

**MapLibre wins on every axis we care about**: OSM continuity preserved (matches BTC Map's aesthetic — Nathan's project also uses MapLibre styles), no vendor lock-in, no API key, true native components.

## Acceptance criteria

- [ ] `@maplibre/maplibre-react-native` added with config plugin for EAS
- [ ] All three map surfaces migrated:
[...]

PR #563, opened 2026-05-16
feat(map): MapLibre Native scaffolding — replace Leaflet-in-WebView (#552)

BenGWeeks:
Closes the umbrella in #552 and #563 — replaces every WebView Leaflet map in the app with native MapLibre via a single shared component.

### One component, nine surfaces

`src/components/LibreMiniMap.tsx` now renders every map in the app:

| Surface | Mode |
|---|---|
| Explore hub mini-map | zoom-only, auto-follow GPS |
| Geo-caches sub-screen (HuntScreen) | zoom-only, auto-follow GPS, legend |
| Places list header map | zoom-only, auto-follow GPS, bbox-filter, legend |
| Full-screen MapScreen | interactive (pan + zoom + recenter), bbox-driven filter, marker-tap → detail sheets, legend |
| Hide-a-Piglet LocationPickerSheet | interactive + crosshair-pick UX |
| EventDetailScreen hero | non-interactive single-pin preview |
| PlaceDetailScreen hero | non-interactive single-pin preview |
| HuntPiggyDetailScreen hero | non-interactive cache pin + user dot at GPS |
| HuntCreateScreen preview | non-interactive picked-pin preview |

Prop surface for the component: `interactive`, `fill`, `crosshair`, `onBoundsChange`, `onOpenLegend`, `onTapMap`, `onSelectMerchant`/`onSelectCache`/`onSelectEvent`, `userLat`/`userLon` (override of camera centre for detail screens), `userAccuracyMetres`.

### Visual polish on top of the swap

- **Pin chassis**: 22 px circle with 1.5 px white border + drop shadow. All pin types share the chassis.
- **Pin glyphs**: Lucide icons (`Zap` for Lightning merchant, `Bitcoin` on-chain, `PiggyBank` Piglet, `MapPin` NIP-GC cache, `Calendar` event). Merchant glyph is derived from BTC Map category via `btcMapIconComponent(m.icon)`, matching the iconography on the Places-for-you rail cards.
- **User dot**: 22 px blue dot + Google-Maps-style pulsing accuracy halo (Animated.View, native driver). Halo diameter is log-scaled to `userAccuracyMetres`; suppressed when accuracy is null.
- **Legend sheet**: every pin-type row uses the same chassis as the map pins (was a mix of 14 px dots and 28 px circles). Drag-down zone extended to include the "Pin types" title strip, not just the 4 px handle pill.
- **BTC Map attribution**: replaced the rasterised Simple Icons asset with `BtcMapLogoSvg` — a `react-native-svg` component rendering the canonical `btcmap.org/images/logo.svg` paths verbatim. Crisp at any size, no antialiasing fringes.

### Performance wins

- **WebView mount eliminated.** Was ~30-36 s on Pixel preview cold start (#560), now sub-second native paint.
- **GrapheneOS Vanadium-sandbox freezing eliminated.** No WebView → no sandbox to freeze.
- **Marker filter on MapScreen is now pure derived state** (`useMemo` over `places` + filter set), no `injectJavaScript` bridge round-trips.
- **Pull-to-refresh** works anywhere on Explore + Geo-caches now that the inline mini-maps are zoom-only and don't compete for the vertical-pan gesture.
- **MapScreen also fetches caches by author**, not just by geohash, so the user's own Piglets surface on the map even when they're physically far from where the Piglet is hidden — mirrors the same pattern Explore already used.

### Code deleted

- `src/components/ExploreMiniMap.tsx` (~720 LOC of inline Leaflet HTML + setup)
- `src/utils/mapWebview/tiles.ts` + `tiles.test.ts`
- `src/utils/mapMeDot.ts`
- `assets/images/btcmap-logo.png`, `assets/images/btcmap-marker.png`
- LocationPickerSheet's `makeHtml` (the WebView body factory)
- MapScreen's `LEAFLET_HTML` constant (~180 LOC) + `webviewRef`, `webviewReady`, `viewportHydrated`, `pendingViewport`, `setViewportInWebView`, `sendMarkers`, `sendCaches`, `recenterOnUser`, `onMessage`, the viewport-hydrate effect, and the two filter-push effects (~280 LOC total)
- `EXPO_PUBLIC_USE_LIBRE_MAP` env flag (was a transitional A/B gate; we now ship MapLibre unconditionally)

Net delta on `MapScreen.tsx` alone: **−230 LOC**.

## Architecture notes

- The native MapLibre view consumes touches before they reach RN's responder tree, so pan/refresh-suppression is handled via the map's own `onRegionWillChange` / `onRegionDidChange` rather than via wrapping-View touch handlers. (An earlier attempt with `onStartShouldSetResponderCapture` set `mapTouched=true` but never cleared it — the responder release event only fires if we actually captured, which we didn't. Documented in the commit history.)
- Inline mini-maps disable `dragPan` so a vertical drag on the map doesn't hijack the outer ScrollView's pull-to-refresh. The full-screen MapScreen enables everything because it owns its viewport.
- `LibreMiniMap` is wrapped in `React.memo`; the parent ExploreHomeScreen uses `useMemo` / `useCallback` on every prop it hands down so the memo's shallow compare actually fires. Dev-mode legend toggle went from 250-1000 ms renders to 50-150 ms.

## Test plan

- [x] `npx eslint <touched files>` — clean (a handful of pre-existing warnings unrelated to this PR)
- [x] AVD (LP_API34, dev client) — Explore + Geo-caches + Places + MapScreen all render; pins iconography matches the legend; legend drag-down dismiss works; map taps open the right detail sheets
- [x] Pixel 8 dev client — same surfaces verified live during the cutover
- [ ] Fresh EAS preview build at this branch tip — Ben's testing flow. Build deferred until merge so we don't churn EAS credits on intermediate states.

## Follow-ups (filed as separate issues)

- #564 — strip dev-mode logging before next prod release
- #565 — re-enable Friends-of-Friends WoT tier behind a foreground compute dialog
- #566 — bisect remaining cold-start + Explore-mount freezes (network-bound, not map-related)
- #567 — NTAG215 PWD/PACK locking + viewable PIN
- #568 — initial MapScreen → MapLibre tracking issue, now fully delivered here
- #569 — pause NWC balance poll when off Home tab
- #570 — re-enable panning on inline mini-maps without breaking pull-to-refresh


BenGWeeks:
# Review: PR #563 — `feat(map): MapLibre Native scaffolding — replace Leaflet-in-WebView` (revised)

*Updated against the rewritten PR description (2026-05-16).*

## Overview

Replaces every WebView Leaflet map with a single native MapLibre component (`LibreMiniMap`) across nine surfaces. 22 files, **+1,195 / −1,578 (net −383)**. The new description now matches the diff faithfully — the "scaffolding-only" framing is gone and the full cutover is documented end-to-end.

## What the updated description does well

- Maps every consumer of the new component to a *mode* (zoom-only / interactive / crosshair / single-pin preview). That's exactly what reviewers need to reason about prop combinations.
- Quantifies the win in real numbers: "WebView mount ~30-36 s on Pixel preview → sub-second native paint" with a backref to #560's instrumentation. That's load-bearing context for the architectural call.
- Documents the **gesture-arbitration story**: native MapLibre swallows touches before they reach RN's responder tree, so the earlier `onStartShouldSetResponderCapture` attempt was structurally wrong, and pan / pull-to-refresh now ride on `onRegionWillChange` / `onRegionDidChange`. Saves the next maintainer hours.
- Documents the **`React.memo` + parent `useMemo`/`useCallback` coupling**: the memo's shallow compare only fires when the parent stops re-creating prop closures. Legend-toggle render drops 250-1000 ms → 50-150 ms.
- Test plan is now checked-off for AVD + Pixel 8 with a deliberately-deferred EAS preview build (avoiding intermediate-state credit churn — sensible).

## Concerns still standing

### Blocking-class

- **Coverage-gate risk.** `src/utils/mapWebview/tiles.test.ts` (~54 lines) is deleted with no replacement. CLAUDE.md scopes coverage to `src/services` / `src/utils` / `src/contexts`, and the deletion will measurably move the line-coverage floor against `main` once the stack lands. Two cheap options: pull the geohash / bbox math the deleted test exercised into a *new* small util under `src/utils/` with the test repointed at it, *or* add equivalent assertions against `decodeGeohash` / `geohashPrefixes` already in `src/utils/geohash.ts` so coverage stays even.
- **OSM tile-server policy.** Inline `OSM_STYLE.sources.osm.tiles = ['https://tile.openstreetmap.org/{z}/{x}/{y}.png']` hits `tile.openstreetmap.org` directly. The OSMF tile-usage policy forbids "heavy use … distributing an app that uses tiles from openstreetmap.org … without prior permission." Lightning Piggy on Play Store / TestFlight arguably qualifies. The previous Leaflet stack had the same problem at lower volume; the migration is the natural moment to swap to MapTiler / Stadia / OpenFreeMap / a self-hosted cache. Worth at minimum a tracking issue alongside #564-#570.

### Moderate

- **SymbolLayer markers before dense-map shipping.** Flat `<Marker>` React children for MapScreen + PlacesScreen will degrade noticeably in dense regions (London, NYC, Berlin) past ~50-100 markers. The description mentions "Lucide icons … merchant glyph derived from BTC Map category" — these are RN-side renders, not GPU sprites. The architectural follow-up (#552 sub-task 1 in the original framing) should land *before* enabling MapScreen in production for users outside a sparse-rural test set. Worth noting in #568 or a new follow-up so it doesn't get lost.
- **`btcMapService.ts` FETCH_TIMEOUT_MS 20 s → 5 s** is bundled in. It's referenced as #566 work in the source comment, but neither the description's "Visual polish" nor "Follow-ups" sections call it out. It's a real behaviour change (faster fallback to cached merchants on slow networks) — give it one line under the "Performance wins" section so the change history is reviewable.
- **`haloDiameter` is pixel-based, not geographic.** `Math.max(50, Math.min(120, 40 + Math.log10(acc) * 22))` caps at 120 px regardless of accuracy — a 1000 m GPS fix and a 50 m fix render at similar visual sizes. For Hunt finders relying on the dot to gauge "am I close enough?", the cap *understates* uncertainty. Either expose a textual "GPS accuracy: ±N m" overlay (cheap, near-term), or follow up with a proper CircleLayer-with-GeoJSON port (right thing, more work) — the comment in `LibreMiniMap.tsx` already flags this as a planned migration.

### Nits

- **`OSM_STYLE` is defined inline in `LibreMiniMap.tsx`.** It works because there's only one consumer, but if any callsite ever wants a different style (Hunt's purple terrain layer? a dark-mode tile pack?) extracting to `src/utils/mapStyle.ts` now is a one-line refactor.
- **`testID` props** on the new recenter / legend / zoom buttons — CLAUDE.md requires them on interactive elements. Quick to add now while the file is hot in your editor.
- **Pin colours (`#7A5CFF` cache, `#5b3aff` event)** are inline hex literals rather than palette tokens. The Copilot review on #520 caught this exact pattern; promoting them to `palettes.ts` (`cachePurple` / `eventViolet`) keeps the palette as the single source of truth for the theme.

## Performance implications

The numbers in the description tell the story (-30 s cold-start mount, sub-second native paint, marker-filter as derived state). Two things worth pinning down before merge:

- **Cold-start delta on a low-end real device** (Pixel 6a-class or older). The current numbers reference Pixel 8 + AVD; a Pixel 6a / Galaxy A-series cold start would confirm whether the win is broad-based or just on flagship hardware.
- **Marker count at zoom 13 in a dense city.** Spot-check on London / Berlin to see whether the flat `<Marker>` rendering is bearable pre-SymbolLayer.

## Test coverage

- `LibreMiniMap` itself is uncovered, defensible under CLAUDE.md's component-test exclusion.
- One existing test deleted (`mapWebview/tiles.test.ts`). See coverage-gate concern above — this is the only mechanical reason a CI cycle could fail.

## Security

- **HTTPS tiles** ✓.
- **No new permissions** — the existing `expo-location` flow stays in place.
- **MapLibre native renders attribution automatically** ✓ — OSM attribution-ToS satisfied; tile-volume ToS is the open question.

## Recommendation

**Approve with the four follow-up asks**, none of which need to gate this merge if filed as tracked issues:

1. Coverage-gate replacement (in-PR — small).
2. OSM tile-server provider follow-up (new issue, before this lands on `main`).
3. SymbolLayer markers before MapScreen ships to broad audience (track in #568 or new).
4. Cold-start delta verification on a Pixel 6a-class device + city-density marker spot-check, as part of the deferred EAS preview build step.

The migration is the right call, the consolidation into one component is excellent, and the description rewrite now genuinely earns its mergeable-state.

---
🤖 Generated by Claude Code (`/review`, revised)

BenGWeeks:
All 9 Copilot inline comments addressed in `a94d2f3`. Full triage table in the commit body. Also folded in the meta-review's blocking-class items (OSM tile-policy → OpenFreeMap, palette tokens for the cache/event hex literals).

PR #573, opened 2026-05-16
fix(hunt): My Piglets friends' finds no longer flicker on cache refresh (#574)

### LINKED ISSUE #574: My Piglets — friends' finds section flickers on every cache-mirror refresh
On the My Piglets page, the Friends' finds section appears, disappears, and reappears repeatedly while the page is open — most visible during cold start and any pull-to-refresh.

## Repro

1. Open My Piglets with an active Trust Graph + at least one friend who's logged a find.
2. Watch the Friends' finds section as relay echoes arrive — it pops in, blanks, and re-fills several times.

## Root cause

`src/screens/MyPigletsScreen.tsx`'s find-log subscribe effects list `cacheByCoord` in their dep arrays and call `setMyFinds(new Map())` / `setFriendFinds(new Map())` at the top of each run. Every change to `allCaches` (relay echo, `mergeCaches` tick, `fetchCachesByAuthor` result) rebuilds `cacheByCoord`, restarts the sub, and wipes the section between renders.

Spotted while testing #567.

BenGWeeks:
## Why

On the My Piglets page, the Friends' finds section pops in, disappears, and reappears repeatedly whenever the relay-sourced cache mirror refreshes — a visible flicker every time a kind 37516 echo or a pull-to-refresh tick lands.

## Root cause

Both find-log subscribe effects in `MyPigletsScreen.tsx` listed `cacheByCoord` in their dep arrays and called `setMyFinds(new Map())` / `setFriendFinds(new Map())` at the top of each run. Every change to `allCaches` (relay echo, `mergeCaches` tick, `fetchCachesByAuthor` result, manual refresh) recomputed `cacheByCoord`, restarted the sub, and wiped the section between renders before the new sub had a chance to repopulate.

## Fix

Stop embedding the matching `ParsedCache` into `FoundEntry`. The event-side fields (id / coord / finderPubkey / createdAt / amountSats) are pure event content — they don't depend on the cache mirror. The render path resolves the cache via `cacheByCoord.get(entry.coord)` on demand, so a late-arriving kind 37516 surfaces the row's name as soon as it lands without restarting the subscription. The friends'-list filter `(e) => e.cache !== null` becomes `cacheByCoord.has(e.coord)` — same intent, recomputed via `useMemo` when the map grows.

## Test plan

1. Open My Piglets while Trust Graph + caches are still warming up. The Friends' finds section should remain stable as the relay sub fills in — no flash of empty / re-populate cycle.
2. Pull-to-refresh several times in quick succession. Section content stays put; entries whose cache hasn't reached the device yet stay hidden until the cache lands, but won't *toggle*.
3. `npx tsc --noEmit` + `npx jest` + `npx eslint src/screens/MyPigletsScreen.tsx` all green.

Closes #574.

### DIFF
--- a/src/screens/MyPigletsScreen.tsx
+++ b/src/screens/MyPigletsScreen.tsx
@@ -447,7 +447,7 @@ const MyPigletsScreen: React.FC<Props> = ({ navigation }) => {
           // Cache resolution moved from the find-log subscribe callback
           // to this render path — see the FoundEntry comment. The lookup
           // is O(1) and we wire `cacheByCoord` into the SectionList via
-          // `extraData` below so a late-arriving kind 37516 always
+          // `extraData` above so a late-arriving kind 37516 always
           // re-runs `renderItem` and the row's name re-paints, even on
           // the my-finds path (whose `foundList` useMemo doesn't itself
           // depend on the cache mirror).
--- a/src/screens/MyPigletsScreen.tsx
+++ b/src/screens/MyPigletsScreen.tsx
@@ -28,10 +28,14 @@ interface Props {
   navigation: ExploreNavigation;
 }
 
-// Parsed kind 7516 found-log keyed by the cache it refers to, kept
-// flat so the SectionList renderer can render a single row regardless
-// of which section it came from. coord is the `<kind>:<pubkey>:<d>`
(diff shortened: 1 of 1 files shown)


BenGWeeks:
@copilot please review

PR #618, opened 2026-05-17
perf(nostr): cache verified-event ids to skip re-Schnorr on re-subscribe

### LINKED ISSUE #31: Allow creating a new Nostr identity (npub/nsec)

BenGWeeks:
## Why

The 2026-05-17 CDP profile of the Explore warm tab-switch freeze (16 s mean, hard JS-thread block) showed **25-30 % of the budget is Schnorr signature verification on the Hermes JS thread** — top frames were `mul` (15.5 %), `add` (4.8 %), `sub` (2.0 %), `sqrtMod` (1.5 %), `rotr`, `invert`, `Maj`, `double`, `update` — all @noble's secp256k1 + SHA-256 primitives. Each kind-37516 cache event coming back from the relay re-subscribe gets `verifyEvent`'d at ~25 ms / event on Hermes (no JIT, slow BigInt). A 50-200 event backfill burst is **1.2-5 s of pure crypto**, blocking the JS thread.

The relay's filter-EOSE replay on every re-subscribe re-emits the same events we already verified on the previous focus. Re-verifying them is wasted work.

This is the first of three remediations:
1. **This PR**: verified-id cache — skip schnorr on already-seen events
2. (separate PR): `skipVerification` for cache/event subs specifically — biggest immediate win
3. (separate PR): native crypto via `react-native-quick-crypto` — broadest win but biggest refactor

## What changes

`src/services/nostrService.ts` — extend the existing `pool.verifyEvent` override (which already fast-paths kind-0) with an id-cache check:

- Module-scoped `Set<string>` + FIFO `string[]` for eviction, capped at **10 000 entries** (~640 KB upper bound, negligible).
- `pool.verifyEvent` checks the cache first; if hit, returns true immediately.
- If miss, runs the original verification (validate-only for kind-0, full schnorr otherwise), and remembers the id *only after* the check passes.
- `__resetVerifiedEventCache` + `__verifiedEventCacheSize` exposed for tests / rare invalidation.

## Security posture

Unchanged. We only cache an id once the full schnorr check passes, and event ids are SHA-256 of the canonical signed encoding — two events with the same id are byte-identical signed events.

## Expected impact

On re-focus of a screen that re-subscribes to the same filter (the common case for Explore tab cycling), the relay's backfill burst will hit the cache for every previously-seen event. **Re-verification path drops to ~0 ms per cached event.** Net: 1-5 s of JS-thread block removed per re-focus, depending on how many of the backfill events were seen on the previous visit.

Cold-start (first ever visit) is unaffected — the first verify is still paid.

## Test plan

Measurement protocol (after merge + install):
```
```

Baseline (pre-PR): mean 16.1 s, p50 16.1 s, p99 16.7 s, max 16.7 s.

Expected post-PR: switch 1 unchanged (cold subs, first verify); switches 2-5 should drop significantly as the cache fills.

## Gates

- ✅ `npx tsc --noEmit`
- ✅ `npx eslint` (one pre-existing array-type warning unrelated to this PR)
- ✅ `npx prettier --check`

Closes #31

### DIFF
--- a/src/screens/account/AboutScreen.tsx
+++ b/src/screens/account/AboutScreen.tsx
@@ -10,6 +10,13 @@ import {
 } from 'react-native';
 import { Image as ExpoImage } from 'expo-image';
 import AsyncStorage from '@react-native-async-storage/async-storage';
+// `expo-file-system/legacy` keeps the `cacheDirectory` string + `getInfoAsync`
+// API. The v55 top-level module switched to the `Paths` / `File` class API;
+// the legacy entry-point is the path of least resistance for a Hermes-profiler
+// dump that just needs a string path to hand to Hermes' native API.
+import * as FileSystem from 'expo-file-system/legacy';
+import * as Sharing from 'expo-sharing';
+import { Alert } from '../../components/BrandedAlert';
 import SecretModeCelebration from '../../components/SecretModeCelebration';
 import * as nip19 from 'nostr-tools/nip19';
 import { UserRound } from 'lucide-react-native';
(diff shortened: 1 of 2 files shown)


BenGWeeks:

Closing as the commits from this PR were rolled into the consolidated #31 perf stack and merged to main as part of #628. Verified-event cache landed via merge commit `2340c71`.

PR #619, opened 2026-05-17
perf(nostr): skip schnorr verify for non-financial kinds (37516 + 31923)

BenGWeeks:
## Why

Second of three remediations for the 25-30 % crypto cost on the Explore tab-switch freeze (#31, 2026-05-17 CDP profile).

PR #618 added a verified-event-id cache that helps re-subscribe paths. This PR goes further: it **skips schnorr verification entirely** for two event kinds where a malicious relay can at worst show fake content without moving funds:

- **kind 37516 (NIP-GC cache listing)** — the LNURL bearer lives on the NFC tag, never on the Nostr event (enforced by `buildCacheListing`). A forged listing puts a fake pin on the map; finder still has to scan a real tag to get sats.
- **kind 31923 (NIP-52 time-based event)** — meetup metadata. A fake "Bitcoin meetup tonight" wastes time, costs no money.

Other kinds keep the full schnorr (DMs, reactions, zap requests/receipts, comments, found logs).

## What changes

`src/services/nostrService.ts` — extend `SKIP_VERIFY_KINDS` set used by the `pool.verifyEvent` override:

```ts
const SKIP_VERIFY_KINDS = new Set<number>([37516, 31923]);
…
if (event.kind === 0 || SKIP_VERIFY_KINDS.has(event.kind)) {
  ok = validateEvent(event);  // cheap structural check, no schnorr
} else {
  ok = verifyEvent(event);    // full schnorr
}
```

## Security posture

Same trade-off `nostr-tools` exposes via `subscribeMany({ skipVerification: true })` — except scoped to event kinds where the threat is bounded. Specifically:

- ✅ Can't exfiltrate sats (no LNURL on the wire)
- ✅ Can't compromise identity (no signing keys involved)
- ⚠️ Can spoof a cache pin / meetup announcement (visible-content forgery)
- ⚠️ Could in theory display a "Hidden by Big Piggy" badge on a hide that wasn't Big Piggy's

The display-only forgeries are filtered downstream by the WoT trust graph anyway — `isTrustedRef.current(c.hiderPubkey)` rejects untrusted pubkeys. So even a forged event has to claim a pubkey the user already trusts to slip through.

## Test plan

Same protocol as PR #618:

```
PIGGY_DEVICE=emulator-5554 PIGGY_APP_ID=com.lightningpiggy.app.preview \
  bash scripts/perf-explore-tab-switch.sh 5
```

[...]

PR #621, opened 2026-05-17
perf(explore): keep relay subs open across tab blur (close #31)

### LINKED ISSUE #31: Allow creating a new Nostr identity (npub/nsec)

BenGWeeks:
## Why

The biggest single lever for the Explore tab-switch freeze.

The pre-existing `useFocusEffect` tore the NIP-GC + NIP-52 relay subscriptions down on **every tab blur** and re-opened them on **every focus**. The cited rationale was *"reconnect on re-focus is ~100 ms; saves JS-thread load while user is on Home"*. **The measurement that prompted that comment was wrong** — the actual cost of re-focus is **15-16 seconds** (#31), because the relay re-emits its full geohash-prefix backlog every time we re-subscribe, the JS thread processes each event through trust-filter + queue + occasional schnorr-verify, and the JS thread blocks while doing so.

The math was upside-down — we were paying a 15-second freeze on every tab return to save the user from <2 ms/min of background event processing while they were on another tab.

## What changes

`useFocusEffect` → `useEffect` for the cache + event subscriptions. The subscription opens once on first mount (the tab navigator's `lazy: true` ensures ExploreHomeScreen doesn't mount until first visit), stays open across blur/focus, only closes on:
- full unmount (logout / navigator reset)
- pull-to-refresh (`refreshKey` bump)

Drops 92 lines of focus-effect machinery (drain-on-blur, cancelled flags' second purpose), keeps the cancelled flag for the unmount-vs-stray-event race we still need.

## Why this is the standard pattern

Every production Nostr client keeps subscriptions open across screen changes:
- **Damus** (iOS) — pool subscribed at app launch, closed only on quit
- **Snort** (web) — pool persistent for the session
- **Amethyst** (Android) — same model

WebSocket stays connected; the relay only sends backfill **once per session**. Tear-down on every screen change was a Lightning-Piggy-specific choice that didn't survive contact with the actual cost.

## Trade-off

Events continue arriving in the background while user is on another tab. After PR #618 (verify-event-id cache) and #619 (skip schnorr for kinds 37516 / 31923) each event costs ~1-2 ms. Typical density of ~10 events/min = ~10-20 ms/min of background JS work. Negligible.

## Test plan

```
```

Baseline (post #618): **mean 15.7 s, p99 15.9 s, 3-7 frames per switch**.

Expected post-PR: switch 1 (first visit, cold backfill) similar; switches 2-5 should drop dramatically — ideally **<1 s each**, since the subscription is already open and no backfill happens.

## Stacks on

- PR #618 (verify-event-id cache)
- PR #619 (skip schnorr for non-financial kinds)

## Gates

- ✅ `npx tsc --noEmit`
- ✅ `npx eslint` (3 pre-existing warnings, none from this PR)
- ✅ `npx prettier --check`

Closes #31

### DIFF
--- a/src/screens/ExploreHomeScreen.tsx
+++ b/src/screens/ExploreHomeScreen.tsx
@@ -542,103 +542,118 @@ const ExploreHomeScreen: React.FC<Props> = ({ navigation }) => {
   const [untrustedCacheCount, setUntrustedCacheCount] = useState(0);
   const [untrustedEventCount, setUntrustedEventCount] = useState(0);
   const subsCloserRef = useRef<(() => void)[]>([]);
-  // NIP-GC + NIP-52 subscriptions are wrapped in useFocusEffect so they
-  // pause on tab blur. Previously the subs stayed open for the rest of
-  // the session once the user visited Explore even once — relay events
-  // kept landing on the JS thread (a Map clone per delivery is small
-  // but not free) while the user was on Home/Messages/Friends, eating
-  // into bridge dispatches for payment-settlement polls + QR-scan
-  // callbacks (#554). Reconnect on re-focus is ~100 ms; foreground
-  // JS-thread responsiveness is the better trade.
-  // Destructure pos into primitives so the focus effect's deps don't
+  // NIP-GC + NIP-52 subscriptions live on the *mount* lifecycle, not
+  // the focus lifecycle. Open once when ExploreHomeScreen first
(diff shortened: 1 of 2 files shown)


BenGWeeks:

Closing as the commits from this PR were rolled into the consolidated #31 perf stack and merged to main as part of #628. Keep-subs-open across tab blur landed via merge commit `edd276b`.

PR #622, opened 2026-05-17
perf(map): pause halo pulse when screen unfocused (closes #31)

### LINKED ISSUE #31: Allow creating a new Nostr identity (npub/nsec)

BenGWeeks:
## TL;DR

**The Explore tab-switch freeze (#31) is caused by `LibreMiniMap`'s halo pulse animation running in the background on every other tab, hammering the JS thread every ~800 ms.** This PR gates the pulse on `useIsFocused()` so it only runs when the host screen is actually visible.

## The diagnosis

Perfetto trace (2026-05-17, three warm Home→Explore switches, 65 s window):

```
62 sleeps × ~815 ms = 50.6 s of trace was JS-thread idle in 815 ms chunks
                    = 78 % of trace time
```

Where did the 815 ms come from? **`LibreMiniMap.tsx:190`**:

```ts
Animated.loop(
  Animated.sequence([
    Animated.timing(pulse, { toValue: 1.18, duration: 800, useNativeDriver: true }),
    Animated.timing(pulse, { toValue: 1.0,  duration: 800, useNativeDriver: true }),
  ]),
).start();
```

The animation itself runs on the UI thread (correct), **but every leg's completion bounces through the JS thread to schedule the next leg** (less well-known RN behaviour for sequence-based loops). Effective JS-thread wake-up every ~800 ms, indefinitely.

This was hiding because:
- Tab navigator's `lazy: true` keeps every visited screen *mounted* — so `LibreMiniMap` stays alive on Home, Messages, Friends
- Per-iteration JS cost is small in isolation (~10–30 ms)
- The existing inline comment confidently said *"native-driven so the JS thread stays free"* — half-true

## Why this matters for the tab-switch freeze

Time-lapse screenshots taken during a tab-switch:

```
t = 0.3 s after tapping Explore: md5 fccee39ca50401de8c34667dcf126ce5  (Home tab content)
t = 1.3 s                       : md5 fccee39ca50401de8c34667dcf126ce5  (same — screen frozen)
t = 2.3 s                       : md5 fccee39ca50401de8c34667dcf126ce5  (same — screen frozen)
t = 4.3 s                       : md5 fccee39ca50401de8c34667dcf126ce5  (same — screen frozen)
t = 9.0 s                       : md5 1d89121c984f9806aa913fe07c9cb42e  (different — Explore now visible)
```

**Screen is genuinely frozen on the previous tab for 4+ seconds after tapping.** The JS thread can't process the tap-event because it's still processing queued pulse-iteration callbacks AHEAD of the tap.

## What changes

`src/components/LibreMiniMap.tsx` — gate the `Animated.loop` on `useIsFocused()`:

```ts
const isFocused = useIsFocused();
useEffect(() => {
  if (!isFocused) {
    pulse.setValue(1);
    return;
  }
  const loop = Animated.loop(...);
  loop.start();
  return () => loop.stop();
}, [isFocused, pulse]);
```

On blur, the loop stops + `pulse` resets to its base value. On refocus, the loop restarts. The halo itself remains painted (the polygon GeoJSON layer doesn't depend on `pulse`); only the *animation* pauses.

## Expected impact

Tap-on-Explore-from-Home should drop from ~4 s "screen frozen" to immediate. Combined with PRs #618 / #619 / #621 (cryptography fixes that together saved ~400 ms), the warm tab-switch should approach the theoretical minimum (~200-500 ms).

## Stacks on

Independent — applies cleanly on top of main *and* on top of #618 + #619 + #621.

## Test plan

```
```

Baseline (with all 3 stacked crypto PRs but without this fix): mean **15,756 ms**, p99 **15,986 ms**.

Expected post-fix: mean **<500 ms**, p99 **<1 s**. If this hypothesis is right, this is the single fix that closes #31.

## Gates

- ✅ `npx tsc --noEmit`
- ✅ `npx eslint`
- ✅ `npx prettier --check`

Closes #31

### DIFF
--- a/src/components/LibreMiniMap.tsx
+++ b/src/components/LibreMiniMap.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useMemo, useRef } from 'react';
 import { Animated, View, StyleSheet, TouchableOpacity, Text } from 'react-native';
+import { useIsFocused } from '@react-navigation/native';
 // Alias MapLibre's `Map` component so we can still use the built-in
 // `Map<K,V>` global for the coord → source lookups below.
 import {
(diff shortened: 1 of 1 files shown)


BenGWeeks:

Closing as the commits from this PR were rolled into the consolidated #31 perf stack and merged to main as part of #628. Pulse-pause when unfocused landed via merge commit `7858f43`.

Issue #627, opened 2026-05-18
feat(hunt): default WoT chip to 'All' for Geo-caches + Events so new users see content


BenGWeeks:
## Problem

Currently the WoT (Web of Trust) chip on the **Explore** page ("Geo-caches near you" + "Events near you" rails) and the dedicated **Geo-caches** + **Events** screens defaults to a Secret-Mode-gated filter (Follows / Trusted). For most users this means **they see nothing** on first launch — the rail is empty / the map has no pins — because none of the published caches or events happen to be from their direct follows.

**Confirmed on the emulator** during the #31 perf testing: a fresh install with no wallet / no Nostr login shows an empty rail even though kind-37516 cache events are arriving from relays. The events are filtered out client-side by the WoT chip default.

A user opening the app for the first time should see *something* — the network's public caches and events, not just their own follow graph.

## Proposal

Default the WoT chip to `'all'` on **all four surfaces**:

- **Explore page** — the inline "Geo-caches near you" rail under the mini-map
- **Explore page** — the inline "Events near you" rail
- **Geo-caches page** (dedicated screen) — same default
- **Events page** (dedicated screen) — same default

Other tabs (Places, Friends, Messages) keep their current defaults — this issue is **only** about the Geo-caches + Events surfaces. Future tuning can introduce a smarter default per-user (e.g. "follows + 1 hop" once the user has follows), but the immediate goal is "new user actually sees content".

## Scope

- Find the WoT-chip initialisation for Geo-caches + Events on all four screens.
- Set initial state to `'all'` instead of whatever Secret-Mode-aware default currently fires.
- User can still toggle to "Follows" / "Trusted" via the chip; only the default changes.
- No change to caching, persistence (if the chip choice persists across sessions, the persisted value should still win — this is just the **first-run** default).

## Notes

Came up while testing the Explore-tab perf fixes — observed an empty Geo-caches rail and empty Events rail on a fresh emulator install even though the rail logic + fetch were healthy. Traced to the WoT default. Worth fixing now while we're touching Explore.

The dev-event deny-list (separate PR) covers the inverse problem: "too many test events from orphaned dev nsecs polluting the rail" — that PR removes leftover noise, this issue makes legitimate events visible by default.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #617, opened 2026-05-17
feat(dev): expose Hermes sampling profiler in About screen (#611 component 1)

BenGWeeks:
## Why

The 2026-05-17 perf audit + this session's tab-switch trace revealed the warm Home→Explore freeze is **"death by 1000 cuts"** — Perfetto shows **1060 small JS-thread Running slices totalling 9 s** during a single 16 s tab switch, not one giant block. Wall-clock metrics + Perfetto traces can pin down *where* the cost is (JS thread, native, blocked) but not *which JS functions* dominate.

To identify the actual hot path we need a **JS function-level flame graph**. Hermes ships a sampling profiler already, but RN doesn't expose it anywhere user-facing — every dev who wants one has to copy-paste boilerplate.

## What changes

`src/screens/account/AboutScreen.tsx` — adds a *Performance profiler (dev)* card with **Start** / **Stop & Share** buttons:

- **Start** → `HermesInternal.enableSamplingProfiler()`.
- **Stop & Share** → dumps via `HermesInternal.dumpSampledTraceToFile(<cache path>)`, disables the profiler, verifies the file landed non-empty, then opens the OS share sheet (expo-sharing) so the dev can save to Files / airdrop / drop straight into **Chrome DevTools → Performance** for the flame graph.

Gated on `__DEV__ || EXPO_PUBLIC_KEEP_PERF_LOGS === '1'` so the card never appears in production. The preview profile in `eas.json` already sets `EXPO_PUBLIC_KEEP_PERF_LOGS=1`, so it surfaces in preview but not production.

## Defensive design

- **No Hermes / no profiler**: `enableSamplingProfiler` is absent → surfaces a "Hermes profiler unavailable" Alert instead of crashing.
- **`disableSamplingProfiler` throws** on some Hermes builds if called without a profile-in-flight → caught silently.
- **Empty profile** (`dumpSampledTraceToFile` no-ops silently sometimes): we `FileSystem.getInfoAsync` the file and surface a "Profile is empty" Alert if size is 0.
- **`expo-file-system/legacy` import**: v55's top-level switched to a Paths/File class API; we want the simple `cacheDirectory` string + `getInfoAsync` for a Hermes API that just needs a path. Comment in source explains.
- **BrandedAlert**: ESLint enforces it per `CLAUDE.md` Code Style.

## Test plan

After install (preview build with `EXPO_PUBLIC_KEEP_PERF_LOGS=1` or any `__DEV__` build):

[...]

PR #628, opened 2026-05-18
perf(explore): kill 12s tab-switch lock — tiered BTC Map fetch + persistent relay subs + skip-schnorr cache (#31)

### LINKED ISSUE #31: Allow creating a new Nostr identity (npub/nsec)

BenGWeeks:
## Summary

Closes #31 — the Explore tab freeze / "tap doesn't register" pattern. Stacks 8 commits + transitively includes PR #617's Hermes profiler dev tool, so merging this PR also closes #618, #619, #621, #622, #617.

**Pixel 8 measurement (real hardware, preview build with this stack):**

| Metric | Before (prod v1.1.0 / vc=67) | After (this stack) | Change |
|---|---|---|---|
| Heartbeat gap > 500 ms during Explore visit | 7,361 ms / 3,782 ms / 4,222 ms | **0** | gone |
| Tab tap → ready (first visit, real hardware) | ~12-22 s | **198 ms** | **~60-100×** |
| `tap → focus` consistency | bursts of 14 queued taps in 5 ms | 119-163 ms every time | smooth |
| Frame jank during Explore session | 48-100% (3 frames / 24 s) | 1.4-2.2% throughout | n/a |
| `fetchPlacesInBbox` per Explore visit | 3 stacked (~10 s each) | 0 (replaced) | gone |
| `fetchCachesByAuthor` | every focus, ~8 s each | once total / 8 switches | n/a |

## What we found

A logcat capture during the user-reproduced 12-second tab lock pinpointed the culprit: a **7,361 ms synchronous JS-thread block** during every `fetchCachesByAuthor` call. The function is async network-wise, but every event the 8 relays sent back went through `verifyEvent` synchronously — `~25 ms` per schnorr check on Hermes, multiplied by 200+ replayed events per resubscribe = ~7 s of solid JS-thread occupation. During that block, tab taps queued on the event loop; when the block ended, 14 queued "Home tap" events fired within 5 ms of each other.

Compounding: a GPS-jitter cascade re-fired the rail `useEffect` on every Android position sample (~1 Hz), even when the user was sitting still — Android returns a fresh `{lat, lon}` object per sample, React compares deps by reference, useEffect fires, spawning a new in-flight fetch that didn't yield once it started processing the 367-place BTC Map response.

## What this PR does

Eight commits, each isolating one rail of the problem:

| Commit | Fix | What it stops |
|---|---|---|
| `824df74` | `fetchNearestPlaces` tiered 25 / 100 / 500 km radii | The ±2° (220 km) BTC Map bbox returning 367 places we never showed |
| `f391f4c` | `posBucket` — round `pos` to 3 decimals (~100 m) via `useMemo` | GPS-jitter cascade re-firing the rail fetch every second |
| `794650e` | `[PerfTab] hidden` + `ready` markers + `perfPageReady` | Future perf work doesn't have to infer tap→usable from gfxinfo |
| `d842b23` | docs: `[PerfTab]` markers, anti-patterns, real-device logcat in `PERFORMANCE.adoc` + 4 new `TROUBLESHOOTING.adoc` rows | Knowledge loss when this freezes again in 6 months |
| `#618` | Cache verified event ids — skip schnorr on re-arrival | The 200+ replayed events per resubscribe each costing 25 ms |
| `#619` | `SKIP_VERIFY_KINDS = [37516, 31923]` — schnorr skip for non-financial kinds | Even uncached cache events take ~0 ms (validate only, no curve math) |
| `#621` | Keep relay subs alive across tab blur | The 8 s `fetchCachesByAuthor` re-firing on every Explore focus |
| `#622` | Pause `LibreMiniMap` halo pulse when screen unfocused | Animated.loop bouncing through JS every 800 ms while user is on Home |

Plus PR #617 (Hermes sampling profiler in About → Performance) which came in transitively from one of the branches — keeping it for the dev tool, scope-flagged in commit `e2b5651`.

## Why bundle vs. ship individually

The 4 perf PRs (#618, #619, #621, #622) each look small in isolation, and earlier in the investigation each one's individual measurement showed no significant win (~400 ms shaved on a 16 s baseline). Stacked, they compose: the cache + skip schnorr removes the per-event verify cost, keep-subs-open removes the recurrence, tiered fetch removes the BTC-Map payload, and posBucket removes the trigger. Removing any one rail leaves the others working but the freeze still reproduces.

## Related

- #627 — follow-up: default Geo-caches WoT chip to "All" (separate UX concern surfaced during testing)
- #335 — closed as superseded earlier in this session (the DM-inbox version of the same anti-pattern fix was already landed by #510 + #508)
- Issue #188 — still open even though main's #510 + #508 resolve it in practice; worth a housekeeping pass

## Test plan

- ✓ `npx tsc --noEmit` clean
- ✓ `npx eslint` — only pre-existing warnings, none introduced
- ✓ `npx prettier --check` clean
- ✓ Real-device test on Pixel 8 (preview build, sideloaded): 8 consecutive Home↔Explore switches, no JS-thread block > 500 ms across the entire sequence, `tap→ready=198ms` on first visit, `tap→focus=119-163ms` on every subsequent visit
- ✓ Emulator test (Pixel 4 AVD, API 35): same shape, `tap→ready=5094ms` on cold first visit (dominated by emulator GPS lock, ~4 s of that)
- ✓ `gfxinfo` healthy throughout — 1.4-2.2% jank on Pixel (was 48-100%)

Closes #31

### DIFF
--- a/app.config.ts
+++ b/app.config.ts
@@ -180,7 +180,7 @@ export default ({ config }: ConfigContext): ExpoConfig => ({
     //   eas build:list --platform android --status finished --limit 1
     //
     // See docs/DEPLOYMENT.adoc → "Local production builds".
-    versionCode: 66,
+    versionCode: 70,
   },
   web: {
     favicon: './assets/favicon.png',
--- a/docs/PERFORMANCE.adoc
+++ b/docs/PERFORMANCE.adoc
@@ -196,6 +196,76 @@ hour, within 5 km, don't hit the network at all. Only use on
 surfaces where the map is decorative; PlacesScreen / MapScreen keep
 the bbox call.
 
+=== `Array.prototype.shift()` as a FIFO eviction on a large array
+
+**Shape:** a bounded cache holds N entries in an array, with a Set
+keyed by the same value for O(1) membership tests. On insert past N,
+the code calls `array.shift()` to evict the oldest entry. This was the
+original implementation of `nostrService.ts`'s `verifiedEventIds` cache
+— PR #628 review flagged it as O(N) per eviction once the cache fills,
+because every remaining element has to be re-indexed.
+
+**Symptom:** the cache itself was meant to *save* JS-thread time
+(skip schnorr re-verification). With `shift()` on a 10 000-entry array,
+the eviction cost ate into the savings — measurable on the same hot
+path the cache is supposed to speed up.
+
+**Fix:** circular buffer + head cursor:
+
+[source,ts]
+----
+const CAP = 10_000;
+const ids = new Set<string>();
+const order = new Array<string | null>(CAP).fill(null);
+let head = 0;
+const remember = (id: string): void => {
+  if (ids.has(id)) return;
+  const evicted = order[head];
+  if (evicted !== null) ids.delete(evicted);
+  order[head] = id;
+  ids.add(id);
+  head = (head + 1) % CAP;
+};
+----
+
+Both insert and evict are O(1). The initial `null` slots evict harmlessly
+during the warm-up window (Set.delete on null is a no-op). Resetting the
+cache is `ids.clear(); order.fill(null); head = 0;`.
+
+**Static test:** `grep -nE '\.shift\(\)' src/` and inspect each — most
+hits in this codebase are legitimate one-shot dequeues; the dangerous
+shape is `.shift()` inside the same function as a matching `.push()`
+when both run on a high-frequency callback.
+
+=== Per-tab Maps keyed by `tabName` for transient state
+
+**Shape:** a `Map<string, T>` keyed by tab name, populated on tap, never
+cleaned. Once the user has visited Home, Messages, Explore, Friends,
+the Map has 4 permanent entries. Any consumer that iterates the Map on
+a frequent event (e.g. an outgoing-tab blur listener that wants to
+emit "tap→hidden") then logs / processes one entry per tab visited.
+
+This was the original `__tabTapAt` in `perfLog.ts` — PR #628 review
+flagged it for accumulating stale entries.
+
+**Fix:** track the most-recent value as a single pair of variables
+rather than a Map. Only one tab transition is ever in flight at a time
+(React Navigation serialises tabPress / focus / blur per gesture), so
+the Map's invariants are stricter than its type — collapse to scalars:
+
+[source,ts]
+----
+let __lastTapAt: number | null = null;
+let __lastTapDestination: string | null = null;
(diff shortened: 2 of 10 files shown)


BenGWeeks:
## Review pass — addressed both Copilot comments + two Stev.ie P2 audit items in one commit (`6621968`)

### Copilot

1. **`perfTabHidden` iterates every entry in `__tabTapAt`** (blocking) — `__tabTapAt` was a Map keyed by tab name that never expired entries (perfTabRendered no longer deleted them, perfTabTap only overwrote the same key), so once the user had visited > 1 tab the Map accumulated permanently and every blur logged one stale `hidden` line per ever-visited tab. **Fix**: collapsed the Map to a single `__lastTapAt` / `__lastTapDestination` pair — only one tab transition is ever in flight at a time, so the Map's invariants were stricter than its type. No leak by construction.

2. **`verifiedEventOrder.shift()` is O(N) on 10k array** (suggestion) — `Array.prototype.shift()` re-indexes every remaining element, eating into the same hot path the schnorr-cache is meant to speed up. **Fix**: circular buffer + head cursor. `new Array<string | null>(CAP).fill(null)` + `head = (head + 1) % CAP`. O(1) insert AND evict. Initial `null` slots evict harmlessly during warm-up (`Set.delete(null)` is a no-op).

### Stev.ie audit P2s

3. **Sort memos still re-ran on every GPS sample** — `sortedMerchants` / `sortedCaches` / `sortedEvents` listed raw `pos` in their dep arrays, so a fresh `{lat, lon, accuracy}` object with identical primitives re-ran three haversine sweeps per GPS tick. Same jitter cascade `posBucket` already closed for the fetch effect. **Fix**: switched to `posLat` / `posLon` primitive deps (already extracted at line 598 for the relay-sub effect).

4. **Pull-to-refresh implicit force** — the SWR short-circuit in `fetchNearestPlaces` was dodged because `refreshDataset()` wiped the cache before the effect re-ran, but the call site didn't say so. **Fix**: `fetchNearestPlaces(pos.lat, pos.lon, 10, { force: refreshKey > 0 })`. Explicit beats implicit, regression-prevention against future `refreshDataset` refactors.

### Docs

`docs/PERFORMANCE.adoc` gains two new "Patterns to avoid" sections:
- `Array.prototype.shift()` as FIFO eviction on a large array (with circular-buffer template)
- Per-tab Maps keyed by tabName for transient state (with scalarisation template)

Both threads resolved via GraphQL.

Issue #695, opened 2026-05-26
perf+security(messages): re-architect the DM store — indexed encrypted SQLite, decrypt-once, persistent sub + since-cursor


BenGWeeks:
## Goal
A genuinely performant + private DM layer, matching the best Nostr clients. Consolidates #687 (cold-start freeze), #689 (interim file-backed cache), #690 (at-rest encryption).

## Peer architectures (source-checked)
- **0xchat** — per-user **Isar** DB, **one indexed row per event**; decrypt **once at receipt in a background isolate**, store **plaintext**; cold start = paginated indexed read, no re-decrypt. Persistent `kind:1059` subscription; incremental `since` rewound **3 days** to absorb NIP-17 random timestamps, cursor advanced off the **inner** event's real time + id-dedup. At rest: **plaintext** (Isar has no encryption) — relies on OS sandbox.
- **Damus** — persist **ciphertext** events in LMDB (nostrdb), **decrypt on demand** into RAM; not app-encrypted (iOS Data Protection).
- **Amethyst** — **in-memory only**, rebuilt from relays each launch; decrypt on demand. `allowBackup=false`.
- **Signal** — message DB **SQLCipher-encrypted**, key wrapped by the Keystore/Keychain.

## Target for LP (0xchat speed + Signal at-rest)
1. **Per-event indexed DB** via `expo-sqlite` (or `op-sqlite`) — rows keyed by `event_id`, indexed `created_at` + `conversation`. Replaces the single-JSON-blob cache (root cause of the 2 MB-row hydrate failure / #687). O(log n) dedupe, paginated reads, no whole-blob parse.
2. **Decrypt once at receipt → store plaintext rows** so cold start is a pure indexed read (no re-decrypt).
3. **Encrypt the store at rest** — SQLCipher (op-sqlite) or app-level AES with a **secure-store/Keystore key** (this is #690). Gives 0xchat speed + Signal confidentiality.
4. **Persistent relay subscription** for kind-1059 + **rewound `since` cursor** (last-seen − ~3 days, advanced off the inner rumor's real `created_at`, dedup by id) so we stop re-querying the whole wrap set each open.
5. **Off-UI-thread decryption** (chunked/yielded or worker) + **pagination** of older messages.

## Notes
- Bigger than #689 (which file-backs the current blob as a stop-gap). This is the strategic target.
- Subsumes #690 (the encryption lands as part of the new store) and the persistent-sub / carry-known-ids perf branches.
[...]

Issue #690, opened 2026-05-26
security(messages): app-level at-rest encryption for the decrypted DM caches (Keystore-held key)


BenGWeeks:
## Background
The `nsec` + LNURLw bearer correctly live in `expo-secure-store` (Android Keystore). But the **decrypted DM caches** — the NIP-17 wrap plaintext cache (file-backed as of #689 / #687), the inbox-summary cache, and the per-conversation caches — sit **unencrypted at the app level** in the private sandbox. They rely only on the Android **app sandbox** + **OS file-based encryption** (FBE).

## Threat model — what FBE alone does NOT cover
FBE protects data only while the device is **off / locked**. Once unlocked, the decrypted message history is readable by:
- root / an exploit / malware with elevated privileges reading `/data/data/<pkg>/`,
- forensic or `adb` extraction on an unlocked (or weakly-locked) device — seizure, 'evil maid', lost-while-unlocked,
- an insecure **backup** (Android auto-backup / `adb backup`) carrying plaintext off-device.

A single such compromise leaks the **entire** decrypted history. This matters more than usual given the GrapheneOS / degoogled, privacy-conscious target audience.

## What Signal does (reference)
Encrypts the message DB itself: **SQLCipher (AES-256)** on Android with the DB passphrase wrapped by a **hardware Keystore** key; Keychain/Secure-Enclave on iOS; `allowBackup=false`. Raw file extraction yields ciphertext.

## Proposed fix (LP equivalent)
1. **Symmetric key in secure-store** — generate a random AES key on first run, store it in `expo-secure-store` (Keystore-backed). Optionally auth-gate it.
2. **Encrypt-on-write / decrypt-on-read** across the DM cache layer — wrap-plaintext cache (file), inbox-summary cache, per-conversation caches. (Need an AES primitive — evaluate `react-native-quick-crypto` / op-sqlite's SQLCipher / a vetted JS AES; the app already bundles `@noble` for hashing.)
3. **Exclude the caches from backup** — `allowBackup=false` or backup-rules exclusion so plaintext (or even ciphertext) can't be swept into a cloud/adb backup.
4. Migration: on first run after the change, re-encrypt or drop+repopulate the existing plaintext caches.

## Scope / notes
- Covers the **whole DM cache layer**, not just the wrap cache from #689.
[...]

PR #697, opened 2026-05-26
chore(storage): SecureStore-backed local DB encryption key (#695 foundation)

BenGWeeks:
First, self-contained step of the encrypted-SQLite local store (#695), which also closes the at-rest gap in #690.

## What
`src/services/localDbKey.ts` — the 256-bit symmetric key that will encrypt the local database at rest (SQLCipher) for DMs and public cached events:
- **`getOrCreateLocalDbKey()`** — generates a random key (`crypto.getRandomValues`, polyfilled) on first call, persists it in **expo-secure-store** (Android Keystore / iOS Keychain, hardware-backed where the device supports it), idempotent, and **single-flight** so two racing cold-start callers can't mint two different keys.
- Persisted writes use `keychainAccessible: AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY`, matching the app's other sensitive SecureStore entries so the key is not included in iCloud / device-migration backups.
- Stored values are validated as 64 hex chars; corrupted values are regenerated and overwritten before use.
- Rejected initialization attempts reset the cached promise so transient SecureStore / IO failures can be retried instead of wedging future callers.
- **`clearLocalDbKey()`** — deletes it on logout / account-wipe; the encrypted DB file is unreadable without it and is deleted alongside (#690).

## Why now / why standalone
Staged delivery of #695. This piece is pure + has **no native dependency**, so it lands + is unit-tested independently of the `op-sqlite`/SQLCipher work (which needs a native rebuild). The next PR opens the encrypted DB with this key.

## Test plan
- 8/8 unit tests (`localDbKey.test.ts`): generation/format, idempotency, persisted-key reuse, single-flight, clear-then-regenerate, invalid persisted-key regeneration, retry after rejected initialization, and SecureStore write options.
- tsc / ESLint / Prettier clean. Services are in unit-coverage scope.


BenGWeeks:
Thanks @copilot — all four addressed in e84e0ff, and the module is renamed `dmDbKey` → `localDbKey` (one encrypted DB will hold DMs **and** public cached events, so it isn't DM-specific):

- **[blocking] keychain backup** — now persists with `keychainAccessible: AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY`, matching `walletStorageService` / `identitiesStore`. No iCloud / device-migration backup.
- **[blocking] cached rejection** — `keyPromise` is reset to `null` on rejection so a transient SecureStore/IO failure can be retried instead of wedging every future call.
- **[suggestion] no validation** — the stored value is now validated as 64 hex chars; a corrupted value is regenerated + overwritten (mirrors `identitiesStore`'s guard).
- **[nit] 'hardware-backed'** — softened to 'hardware-backed where the device supports it'.

Test count went 5 → 8 to cover the new paths.

BenGWeeks:
Re-review nits addressed in 9d36507:
- `SECURE_OPTIONS` now only on `setItemAsync` (dropped from get/delete — `keychainAccessible` only affects writes; matches `walletStorageService`/`identitiesStore`).
- Doc: corrected the peer-client claim — Damus stores events unencrypted, Amethyst is in-memory, 0xchat's Isar is plaintext; we encrypt at rest (Signal's model).
- Doc: clarified the `localDbKey.ts` cross-reference (first slice of #695/#690, added in #697).

PR #698, opened 2026-05-26
fix(hunt): apply dev-leftover denylist on cache storage hydrate + save (#699)

### LINKED ISSUE #699: Orphaned "Geo-Cache 1" stashes still painted from local storage despite denylist
Three **"Geo-Cache 1"** stashes (`d=big-piggy-geo-cache-1`) still show in Explore / Geo-caches on a real device even though their four disposable signers are all on `devEventDenylist.ts`.

**Root cause:** the denylist is enforced only at the relay-ingestion layer (`nostrPlacesPublisher`). `nostrPlacesStorage` rehydrates `ParsedCache`s from its persisted AsyncStorage blob with no `isDevLeftover` check, so events persisted before the signer was denylisted survive and paint on cold start.

**Fix:** re-apply `isDevLeftover` in `nostrPlacesStorage` on hydrate (purge stored) and on save (never re-persist), for both caches and events.

Fixed by #698.

BenGWeeks:
## Problem

Three orphaned **"Geo-Cache 1"** stashes (`d=big-piggy-geo-cache-1`) still appear in the app even though their four disposable signers are all on the `devEventDenylist`.

The denylist was enforced **only at the relay-ingestion layer** (`nostrPlacesPublisher` — the `subscribeMany`/`fetchCachesByAuthor`/`fetchCache` paths). But `nostrPlacesStorage` rehydrates `ParsedCache`s straight from a persisted AsyncStorage blob with **no `isDevLeftover` check**. Events persisted *before* a signer was denylisted survive in that blob and get painted on cold start — the ingestion filter only stops *new* events reaching storage, it never purges what's already there.

Confirmed on-relay: exactly the 4 denylisted pubkeys publish `name="Geo-Cache 1"`, no missing entries — so the list is complete; the leak is the storage layer.

## Fix

Re-apply `isDevLeftover` in `nostrPlacesStorage` at both ends:
- **hydrate (read)** — filters denylisted entries out of the on-disk blob, so the stale "Geo-Cache 1" rows vanish on the next cold start.
- **saveCaches / saveEvents (write)** — never re-persist a denylisted signer.

Covers both caches (`hiderPubkey`) and events (`organiserPubkey`).

## Test plan

- `npx jest src/services/nostrPlacesStorage.test.ts` — 5 unit tests: hydrate-drop + save-drop for both caches and events, plus a clean-entries-untouched guard. All green.
- `npx tsc --noEmit` — clean.
- On device: cold-start the app and confirm the three "Geo-Cache 1" stashes (`d=big-piggy-geo-cache-1`) no longer appear in Explore / Geo-caches.

Closes #699

### DIFF
--- a/src/services/nostrPlacesStorage.test.ts
+++ b/src/services/nostrPlacesStorage.test.ts
@@ -111,4 +111,24 @@ describe('nostrPlacesStorage denylist enforcement (#699)', () => {
     await seedAndRehydrate(CACHES_STORAGE_KEY, [cache(CLEAN, 'A'), cache(CLEAN, 'B')]);
     expect((await loadCachedCaches()).length).toBe(2);
   });
+
+  it('rewrites the sanitized blob to disk when hydrate drops entries (no re-filter every start)', async () => {
+    await seedAndRehydrate(CACHES_STORAGE_KEY, [
+      cache(DENYLISTED, 'Geo-Cache 1'),
+      cache(CLEAN, 'Keeper'),
+    ]);
+    await loadCachedCaches();
+    // The on-disk blob itself should now be clean, not just the in-memory mirror.
+    const raw = await AsyncStorage.getItem(CACHES_STORAGE_KEY);
+    expect(raw).not.toContain(DENYLISTED);
+  });
+
+  it('does not rewrite the blob when nothing is dropped', async () => {
+    await seedAndRehydrate(CACHES_STORAGE_KEY, [cache(CLEAN, 'A')]);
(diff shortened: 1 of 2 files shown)


BenGWeeks:
Good catch @copilot — fixed: when `hydrate()` drops any denylisted entries it now rewrites the sanitized blob (fire-and-forget), so the on-disk copy is actually cleaned instead of re-filtered every cold start. Guarded on a drop count so a clean blob is never needlessly rewritten. Added two tests (rewrite-on-drop, no-write-when-clean).

PR #700, opened 2026-05-26
chore(storage): op-sqlite + SQLCipher encrypted DB module (#695)

BenGWeeks:
## What

#695 step 1: the encrypted local database that the freeze-killing data-access layer will sit on.

- Adds `@op-engineering/op-sqlite` (16.1.1) with **SQLCipher** enabled as the compile target (`package.json` → `"op-sqlite": { "sqlcipher": true }`).
- `src/services/localDb.ts` — opens the single encrypted DB with the Keystore key from `localDbKey.ts` (#697), migrates **schema v1** (`dm_messages`, indexed by `(conversation, created_at)` for the paginated slice reads that replace the blob cache), single-flight + retry-on-failure.
- `verifyEncryptedDb()` — the #695 step-0 **spike smoke-check**: confirms `PRAGMA cipher_version` is non-empty and an encrypted row round-trips.

## Why

The Messages-tab freeze is `refreshDmInbox` doing O(whole-inbox) decrypt synchronously on the JS thread (measured **52s / 49s frozen** on-device, 2026-05-26). Indexed encrypted rows + decrypt-once is the fix; this PR lays the encrypted store. Architecture: `docs/DATA_STORAGE.adoc`.

## Test plan

- [x] `npx jest src/services/localDb.test.ts` — 6 tests (open-with-key, schema, single-flight, retry, verify pass/fail) against a **mocked** native module. `tsc`/eslint/prettier clean.
- [x] **Native rebuild verified on-device** (`npx expo run:android` → LP_API34 emulator): `BUILD SUCCESSFUL`, SQLCipher CMake target compiled, and a dev-only `verifyEncryptedDb()` call logged:
  ```
  [SQLCipher-spike] OK — cipher_version=4.14.0 community
  ```
  → encrypted open + schema migrate + encrypted row round-trip all succeed. The temp wiring was reverted (not in this PR).
- [x] Dev-client build boots with the new-arch native module added.

Part of #695 (1 of ~5 PRs). Next: the DM data-access layer + decrypt-once ingest that actually removes the 52s `refreshDmInbox` freeze.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@gorhom/bottom-sheet": "^5.2.8",
         "@maplibre/maplibre-react-native": "^11.2.1",
         "@noble/hashes": "^2.0.1",
+        "@op-engineering/op-sqlite": "^16.1.1",
         "@react-native-async-storage/async-storage": "2.2.0",
         "@react-native-community/netinfo": "^12.0.1",
         "@react-navigation/bottom-tabs": "^7.15.8",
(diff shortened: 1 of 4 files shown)


BenGWeeks:
All three addressed in be67574:
- **[blocking] silent plaintext risk** — `openLocalDb` now runs `PRAGMA cipher_version` right after open and **throws if it's empty**, refusing to open a plain-SQLite DB. If SQLCipher ever drops out of the build, we fail loud instead of silently writing private DM/transaction rows in cleartext. (Verified on-device it reports `4.14.0 community`.)
- **[suggestion] smoke-row cleanup** — `verifyEncryptedDb` now deletes the `__smoke__` row in a `finally`.
- **[suggestion] dev-only guard** — `verifyEncryptedDb` is now `__DEV__`-gated (throws in production); the permanent SQLCipher assertion lives in `openLocalDb`, so the diagnostic export can't do harm.

+2 unit tests (open-time guard rejects empty cipher; cipher check precedes the schema).

PR #702, opened 2026-05-26
docs(storage): Mermaid DB schema + storage-map diagrams (#695)

BenGWeeks:
Adds three Mermaid diagrams to `docs/DATA_STORAGE.adoc`, capturing the schema design worked out in discussion:

1. **Encrypted DB schema** (`erDiagram`) — target tables: `dm_messages`, `group_messages`, `transactions`, `swaps`, `profiles`, `caches`, `events`, with soft pubkey relationships (Nostr identities, not enforced FKs).
2. **Storage map** (`flowchart`) — the decision rule: secret → Keystore · sensitive/growing collection → encrypted DB · scalar setting → AsyncStorage · public external bulk dataset → plain cache.
3. **Migration** (`flowchart`) — populate the encrypted DB from canonical sources (relays / wallet / BTC Map) and **delete** the old plaintext caches rather than import them.

Records the decisions made while designing #695:
- transactions + Boltz **swaps** are sensitive → their own tables in the encrypted DB (swaps are recovery-critical).
- BTC Map **places** are public external bulk data → deliberately a **plain cache**, not in the encrypted user DB (keeps logout = clean user-data wipe).
- migration is **relays → decrypt-once → DB**, not old-store → new-store.

## Test plan

N/A — docs only. Mermaid blocks are valid (eyeballed; `mmdc` render needs a browser unavailable in CI). They render in the asciidoctor-diagram/PDF pipeline and Mermaid-aware viewers; GitHub's raw `.adoc` preview shows them as source (noted inline).

Part of #695.

### DIFF
--- a/docs/DATA_STORAGE.adoc
+++ b/docs/DATA_STORAGE.adoc
@@ -177,7 +177,7 @@ erDiagram
         TEXT content
     }
     transactions {
-        TEXT id PK
+        TEXT id PK "payment_hash | txid"
         TEXT wallet_id
         TEXT direction
         INTEGER amount_sats
@@ -217,8 +217,8 @@ erDiagram
     }
     profiles ||--o{ dm_messages : "sender"
     profiles ||--o{ group_messages : "sender"
-    profiles ||--o{ caches : "hider"
-    profiles ||--o{ events : "organiser"
+    profiles ||--o{ caches : "hider_pubkey"
+    profiles ||--o{ events : "organiser_pubkey"
 ....
 
 `dm_messages` ships first (#695); the rest migrate in later slices (DMs cause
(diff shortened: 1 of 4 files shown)


BenGWeeks:
All four addressed:
- **[blocking] places contradiction** — fixed. The "What goes in the DB" table no longer lists BTC Map places; it now adds transactions + swaps, and a paragraph states places are the deliberate plain-cache exception (kept out so logout = clean user-data wipe). Doc is internally consistent now.
- **[suggestion] PDF rendering** — now true: `generate-docs.sh` wires `-r asciidoctor-diagram` + `puppeteer-mermaid.json`, `DATA_STORAGE.adoc` is `include`d in the technical-solution PDF, and I verified all three diagrams render to images. NOTE wording updated.
- **[nit] memoised → memoized** — done.
- **[suggestion] 🗑 emoji** — replaced with a plain `discarded` node (font-safe in the Helvetica/DejaVu PDF theme).

BenGWeeks:
Re-review addressed:
- **[blocking] missing dep check** — `generate-docs.sh` now verifies `asciidoctor-diagram` (gem) **and** `mmdc` (mermaid CLI) before building, each with an install hint, so a missing dep fails actionably instead of producing an unrendered/aborted PDF.
- **[suggestion] transactions key mismatch** — ER diagram's `transactions.id` is now annotated `"payment_hash | txid"`, matching the table.
- **[nit] relationship labels** — ER now labels the caches/events relationships with the real column names (`hider_pubkey`, `organiser_pubkey`).

PR #705, opened 2026-05-26
chore(storage): decrypt-once DM ingest + inbox read layer (#695)

BenGWeeks:
#695 steps 3a + 3b-i — the decrypt-once ingest and the inbox read seam that `NostrContext` will delegate to. **Pure + fully unit-tested; not yet wired into the live path** (that's 3b-ii, with on-device before/after).

## What

- **`dmIngest.ts`** — `ingestWraps(wraps, decrypt)`: checks `selectKnownEventIds` (one indexed query) and decrypts **only** wraps not already stored, batched-upserts the fresh rows, and **yields to the JS thread every N fresh decrypts** so a large first-sync stays interactive instead of freezing in one ~49s block. Cooperative abort + progress callback.
- **`dmInbox.ts`** — `loadInboxEntries()` (latest-per-conversation, the indexed read that replaces the whole-blob parse) + `loadConversationEntries()` (paginated) + the pure `rowsToInboxEntries` projection.
- **Schema:** `dm_messages` gains `from_me` + `wire_kind`. Extracting the inbox mapping surfaced that `DmInboxEntry` needs `fromMe` + `wireKind` (the latter is load-bearing for the NIP-04/NIP-17 dedup in `buildDmSummaries`); storing them keeps reads pure (no pubkey threading / kind assumption).

## Why

The Messages-tab freeze is `refreshDmInbox` re-walking + re-decrypting the whole inbox synchronously (52s / 49s frozen, measured on-device). This layer makes the inbox an indexed, decrypt-once, off-thread read. See `docs/DATA_STORAGE.adoc`.

## Test plan

- [x] 26 unit tests across `dmDb` / `dmIngest` / `dmInbox` / `localDb` (decrypt-once gate, yield/abort, projection, schema round-trip). `tsc`/eslint/prettier clean; file-size gate green.
- [ ] **3b-ii (next):** wire `NostrContext.refreshDmInbox`/`fetchConversation` onto `ingestWraps` + `loadInboxEntries`/`loadConversationEntries`, delete the old plaintext wrap cache, and verify on emulator: `perf-messages-tab-open` + `perf-dm-refresh-explore-mount` before/after, plus `test-nip17-send-1on1` / `test-dm-send-amber-373` / `test-multi-account-cold-start-isolation`.

Part of #695. The live rewire (3b-ii) lands separately so the high-stakes DM-path diff gets focused review + measurement.

### DIFF
--- a/src/services/dmDb.test.ts
+++ b/src/services/dmDb.test.ts
@@ -28,6 +28,8 @@ const row = (over: Partial<DmMessageRow> = {}): DmMessageRow => ({
   createdAt: 100,
   sender: 's1',
   content: 'hi',
+  fromMe: false,
+  wireKind: 14,
   ...over,
 });
 
@@ -75,7 +77,7 @@ describe('dmDb', () => {
       expect(mockExecute).toHaveBeenCalledTimes(2);
       const [sql, params] = mockExecute.mock.calls[0];
       expect(sql).toContain('INSERT OR REPLACE INTO dm_messages');
-      expect(params).toEqual(['e1', 'convA', 100, 's1', 'hi']);
+      expect(params).toEqual(['e1', 'convA', 100, 's1', 'hi', 0, 14]);
     });
   });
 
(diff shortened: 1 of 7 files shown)


BenGWeeks:
Both blocking points fixed in the latest push:
- **[blocking] schema migration** — `CREATE TABLE IF NOT EXISTS` indeed won't add `from_me`/`wire_kind` to a pre-existing `dm_messages` (the dev-build table is exactly this case). Added `migrateDmMessagesColumns`: `PRAGMA table_info` → `ALTER TABLE ADD COLUMN` for any missing column on open (idempotent, preserves synced rows).
- **[blocking] decryptor throw** — `ingestWraps` now wraps the `decrypt(wrap)` call in try/catch and treats a throw as undecryptable, so one bad wrap can't abort the whole sync (matches the `WrapDecryptor` docstring). Added a test asserting it keeps going + counts it.

Issue #707, opened 2026-05-26
Extract NostrContext provider into custom hooks to reach the 1,000-line cap


BenGWeeks:
Follow-up to the module-scope extraction (#706, which took `NostrContext.tsx` 4,279 → 3,565). The remaining bulk is the React provider + its ~63 interdependent hooks. To get the file **under the 1,000-line cap** (#703, #704), extract the provider's domain clusters into custom hooks, **one per PR**, each re-verified by the full jest suite (these are stateful closures sharing refs/state, so the cut is delicate — not a one-shot).

## Clusters to extract (rough line estimates)

| Custom hook (new file) | ~lines | Status |
|---|---|---|
| `useDmInbox` (inbox + conversations) | ~900 | **being done now** (separate PR; the keystone — also where the #695 freeze rewire lands) |
| `useProfiles` (kind-0 fetch/cache) | ~400 | this issue |
| `useContacts` (kind-3 follows) | ~300 | this issue |
| `useRelays` (relay list) | ~300 | this issue |
| `useGroups` (group messages/membership) | ~350 | this issue |
| `useNostrSession` (login/logout/signer) | ~250 | this issue |
| `NostrContext` left as a thin composer | ~400–600 | the target |

## Approach (per cluster)
- Move the cluster's `useState`/`useRef`/`useEffect`/`useCallback` into `src/contexts/use<Cluster>.ts`, threading shared deps as hook args / return values.
- Provider composes: `const profiles = useProfiles(session); …` then builds the context value.
- Behavior-preserving — full jest suite must stay green; `tsc`/eslint/prettier/file-size gates pass; ratchet the `NostrContext.tsx` baseline down in `scripts/check-file-size.sh` after each.

The `useDmInbox` extraction is handled separately (it's tied to the #695 DM-store rewire). This issue tracks the other five clusters.

Part of #703.

PR #706, opened 2026-05-26
chore(nostr): extract NostrContext module-scope helpers into sibling modules

BenGWeeks:
First step toward getting `NostrContext.tsx` under the 1,000-line cap (#703, #704). **Pure structural extraction — zero behavior change.**

## What

Moves `NostrContext.tsx`'s ~990 lines of *module-scope* helpers (everything above the React provider) into 8 focused sibling modules, split by concern. The provider + all its hooks are **untouched** (hook count 63 → 63).

`NostrContext.tsx`: **4,279 → 3,565** (−714). New modules (all well under cap):

| Module | Lines | What |
|---|---|---|
| `nostrDmCache.ts` | 306 | NIP-17 wrap-cache + DM inbox/conversation cache + merges |
| `nostrDecryptPacing.ts` | 157 | yield scheduler + decrypt frame-budget constants |
| `nostrGroupRouting.ts` | 125 | `tryRouteGroupRumor` + group route types |
| `nostrEventBus.ts` | 50 | DM/group notify+subscribe pub-sub |
| `nostrCacheKeys.ts` | 50 | profile/contacts/relay cache keys + `readCachedWithTtl` |
| `nostrSecretKeyCache.ts` | 48 | memoised secret-key cache |
| `nostrContextTypes.ts` | 33 | `SignedEvent` / `ConversationMessage` / `RefreshDmInboxOptions` |
| `nostrAuthKeys.ts` | 3 | nsec/pubkey/signer-type storage keys (shared leaf, avoids a cycle) |

Externally-imported symbols are **re-exported** from `NostrContext.tsx`, so no other importer files changed — blast radius is one modified file + the 8 new ones. Ratchets the file-size baseline 4,279 → 3,565.

## Verification (independently re-run, not just the agent's word)

- `tsc --noEmit` clean
- `jest`: **700/700 tests pass** (57 suites) — identical to main → behavior preserved
- `eslint .`: 0 errors (111 warnings = main baseline)
- `prettier --check` clean; `check-file-size.sh` passes
- diff: 9 files; component/hooks region byte-identical (63 hooks → 63)

## Test plan

- [x] Full jest suite green (behaviour preservation).
- [x] tsc / eslint / prettier / file-size gates green.
- [ ] (Optional) emulator smoke — pure refactor, no runtime change expected; covered by the suite + the next DM-path PR's on-device run.

Part of #703 (the next slices extract the component's hook-clusters — DM via 3b-ii — toward the cap).

[...]

Issue #703, opened 2026-05-26
Break up source files over 1,000 lines (split by concern)


BenGWeeks:
Per the new **File size and modularity** rule in `CLAUDE.md` (no source file over 1,000 lines, split by concern), 13 files are over the cap and should be broken up — opportunistically when touched, worst-first:

- 4,279 `src/contexts/NostrContext.tsx` → useDmInbox/useProfiles/useRelays hooks + src/services/dm* (the #695 ingest work already cracks this open)
- 3,121 `src/screens/HuntCreateScreen.tsx` → wizard steps to components + form hook
- 2,173 `src/contexts/WalletContext.tsx`
- 1,710 `src/screens/HuntPiggyDetailScreen.tsx`
- 1,562 `src/screens/MapScreen.tsx`
- 1,529 `src/services/nostrService.ts` → nostrRelay/nostrDm/nostrProfile
- 1,455 `src/screens/ConversationScreen.tsx`
- 1,418 `src/components/TransferSheet.tsx`
- 1,377 `src/screens/ExploreHomeScreen.tsx`
- 1,242 `src/services/nfcService.ts`
- 1,176 `src/components/SendSheet.tsx`
- 1,015 `src/screens/GroupConversationScreen.tsx`
- 1,009 `src/services/nwcService.ts`

Optional follow-up: a grandfather-aware CI gate that fails on *new* over-cap files while letting this list shrink.

PR #708, opened 2026-05-26
chore(nostr): extract DM inbox + conversation into useDmInbox hook

BenGWeeks:
**Stacked on #706** (retargets to `main` once #706 merges). The keystone hook-cluster extraction toward the 1,000-line cap — and the file where the #695 freeze rewire will land. **Behavior-preserving — zero logic change.**

## What

Lifts the DM-inbox + conversation cluster out of `NostrProvider`. The cluster is ~1,500 dense lines, so to keep every file under cap it's split into a hook + two helper modules:

| File | Lines | What |
|---|---|---|
| `useDmInbox.ts` | 880 | state/refs + `refreshDmInbox` + signer decryptors + cache hydrate + live-sub wiring + lifecycle effects |
| `nostrLiveDmSub.ts` | 516 | live kind-4/1059 subscription body (`startLiveDmSubscription`) |
| `nostrFetchConversation.ts` | 384 | thread-open fetch (`fetchConversationFor`) |

`NostrContext.tsx`: **3,565 → 2,031** (−1,534). Combined with #706: **4,279 → 2,031**.

The hook takes `{ pubkey, isLoggedIn, signerType, followPubkeys, getReadRelays }` and returns the context-surfaced values + lifecycle handles. **The context value object is byte-for-byte identical** (diff-verified — same keys, values, dep array). Mid-flight identity guards (`viewerPubkey !== pubkey`) preserved verbatim.

## Why

With the DM logic isolated in an ~880-line `useDmInbox.ts`, the #695 freeze rewire (wiring onto `dmIngest`/`dmDb` from #705) happens in a focused file instead of the 3,565-line context.

## Verification (independently re-run)

- `tsc` clean · `jest` **700/700** (= baseline → behavior preserved) · `eslint` 0 errors · `prettier` clean · `check-file-size.sh` passes (baseline ratcheted 3,565 → 2,031)
- diff: 5 files; context value keys identical

## Test plan

- [x] Full jest suite green (behaviour-preservation proof); tsc/eslint/prettier/file-size gates green; context value diff-identical.
- [ ] Covered on-device by the upcoming #695 DM-store rewire's emulator run (no runtime behaviour change here).

Progresses #703 / #707.


BenGWeeks:
Thanks @copilot. Both relate to **behavior-preserving extraction** scope:
- **[blocking] abort/TTL semantics (useDmInbox.ts:742)** — this is *pre-existing* behavior, present verbatim in `NostrContext` before this PR (multiple `if (signal?.aborted) return;` sites on `origin/main`). The extraction moved it unchanged (byte-identical context value, 700/700 tests). Changing abort/TTL semantics here would be an untested behavior change to the live message path, out of scope for a pure refactor. Filed as #709 to fix deliberately during the #695 3b-ii rewrite, which rewrites this exact function.
- **[nit] React import type (nostrLiveDmSub.ts:1)** — fair; behavior-neutral. This file is rewritten in 3b-ii (the freeze rewire lands here), so I'll switch to `import type` then rather than spin another CI cycle on the refactor PR.

Resolving both as out-of-scope-for-this-refactor / tracked.

Issue #710, opened 2026-05-26
Security follow-ups from #695 at-rest-encryption review (wire into 3b-ii / future)


BenGWeeks:
From the security review of the merged #695 at-rest encryption work. **Verdict: design sound, crypto correct, no Critical findings, no live data exposure today** (the DB layer isn't wired into the live DM path yet). Real CSPRNG key in the OS keystore with `AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY`, working fail-loud SQLCipher guard (runs before any write, unit-tested), fully parameterized DAL (no injection), decrypt-once ingest. Action items below are for the upcoming wiring — capturing so they don't slip.

### H1 (High) — wire `clearLocalDbKey` + encrypted-DB-file deletion into logout/wipe
`clearLocalDbKey` (`localDbKey.ts`) has **zero production callers**, and `wipeAccountCaches` (NostrContext logout) deletes the old caches but **not** the keystore key or `lightningpiggy.db`. Harmless today (DB is empty), but the moment DM ingest writes real rows, a logout would leave the encrypted DB **and** its valid key both on disk → defeats "logout = clean wipe". **Must land in lockstep with the first real DM write (3b-ii / #709).** Add a test asserting logout removes both key + DB file.

### M1/M2 (Medium) — migration must actually delete the old plaintext caches, verified
The #690 plaintext-gap closure isn't implemented; decrypted DM content is still written cleartext to the file-backed wrap cache + AsyncStorage inbox/conversation blobs (the current live store). When the relays→DB migration lands, the post-migration delete of those plaintext caches must (a) be part of the same change and (b) be **verified** (`f.exists` after delete) rather than fire-and-forget `.catch(()=>{})` — don't advance a "migrated" flag until the plaintext source is confirmed gone.

### L1 (Low) — corrupt-key regeneration is unsafe once recovery-critical data lands
`resolveKey` regenerates the key if the stored one fails the hex check. Fine while the DB holds rebuildable-from-relays *cache* data, but **transactions/swaps are recovery-critical** (can't re-fetch) — a regenerated key would orphan them. Before those tables land: either separately-keyed store, or the corrupt-key path must not discard recovery-critical data. (Add a code comment noting the regenerate path is only safe for cache data.)

Refs #695, #709. The crypto in the merged code needs no changes.

BenGWeeks:
Closing — all three action items from the #695 at-rest review are now addressed on main:

[...]

Issue #709, opened 2026-05-26
Revisit refreshDmInbox abort semantics (aborted run still bumps TTL marker)


BenGWeeks:
Surfaced by Copilot on #708. `refreshDmInbox`'s outer wrapper updates `dmInboxLastRefreshAt` (the TTL marker) even when the run was aborted via its `AbortSignal` (the inner task does `if (signal?.aborted) return;` and resolves rather than rejecting). An aborted focus-effect/pull-to-refresh run could therefore bump the TTL and cause the *next* refresh to be skipped as 'too recent'.

**This is pre-existing behavior** (present verbatim in NostrContext before the #708 extraction — see the multiple `if (signal?.aborted) return;` sites), so it was deliberately preserved by the behavior-preserving extraction, not introduced. 

Best evaluated/fixed during the #695 3b-ii rewrite, which rewrites `refreshDmInbox` onto `dmIngest`/`dmDb` anyway — at which point the correct abort/TTL contract (throw AbortError + skip the TTL bump on abort) can be designed + tested deliberately. Part of #695.

BenGWeeks:
Addressed by #733 (defer inbox refresh + **write a short TTL on abort**, breaking the fast-tab-hop chaining trap). #733 fixes this on the current architecture; the #695 3b-ii rewrite will carry the same abort/TTL contract into the DB-backed path. Close this when #733 merges.

PR #712, opened 2026-05-26
chore(storage): logout-wipe primitives for the encrypted DM store (#710)

BenGWeeks:
Addresses the implementable parts of the #695 at-rest-encryption **security review** (#710). The review's verdict was *crypto sound, no live exposure* — these are the hardening primitives for the not-yet-wired logout path.

## What

- **H1 primitive** — `clearLocalDb()`: closes + deletes the encrypted DB file (`op-sqlite` `db.delete()`) and resets the open handle (opens a bare handle to delete the file if the DB wasn't opened this session). Plus **`wipeLocalDmStore()`** = `clearLocalDb()` + `clearLocalDbKey()`, so logout wipes the ciphertext file **and** its keystore key in one call.
- **L1 guard** — comment in `localDbKey.resolveKey` flagging that regenerate-on-corrupt is only safe while the DB holds rebuildable cache data (revisit before recovery-critical transactions/swaps).

## Deliberately deferred (to #709 / 3b-ii)
- **H1 wiring** — calling `wipeLocalDmStore()` from `wipeAccountCaches`/logout. Lands *in lockstep with the first real DM write* (the review's explicit guidance) — and `NostrContext.tsx` is at its file-size baseline (2,031), so the gate won't let me grow it here anyway.
- **M1/M2** — the migration's plaintext-cache deletion + verify-not-silent. The migration doesn't exist until the rewire.

So today both halves are no-ops (nothing's been written to the DB) — this just makes #709's wiring a one-liner and the guarantee correct from the moment ingest lands.

## Test plan

- [x] `npx jest src/services/localDb.test.ts` — 10 tests incl. clearLocalDb (open + bare-handle cases) and wipeLocalDmStore (deletes DB + clears key). tsc/eslint/prettier/file-size green.
- [ ] On-device `db.delete()` behaviour covered when logout wiring lands in #709.

Part of #710. Keeps #710 open for H1-wiring + M1/M2.

### DIFF
--- a/src/services/localDb.test.ts
+++ b/src/services/localDb.test.ts
@@ -91,23 +91,21 @@ describe('localDb', () => {
     expect(cipherIdx).toBeLessThan(schemaIdx); // guard runs before schema
   });
 
-  it('clearLocalDb deletes the encrypted DB file (open handle case)', async () => {
-    const { getLocalDb: g, clearLocalDb } = require('./localDb');
-    await g(); // open it this session
-    await clearLocalDb();
-    expect(mockDelete).toHaveBeenCalled();
-  });
-
-  it('clearLocalDb opens a bare handle to delete when not opened this session', async () => {
-    const { clearLocalDb } = require('./localDb');
-    await clearLocalDb(); // never opened
-    expect(mockOpen).toHaveBeenCalledWith(expect.objectContaining({ name: 'lightningpiggy.db' }));
-    expect(mockDelete).toHaveBeenCalled();
-  });
-
-  it('wipeLocalDmStore deletes the DB AND clears the keystore key (#710 H1)', async () => {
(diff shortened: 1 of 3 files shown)


BenGWeeks:
Good catch — fixed via your option (1): `clearLocalDb` is now **module-private**, so the only public wipe is `wipeLocalDmStore` (which deletes the file **and** clears the key together). That removes the footgun where a standalone `clearLocalDb` could silently leave a readable encrypted DB on a failed delete. I also replaced the misleading 'the key removes it anyway' justification: a delete failure is now a `__DEV__` warn (a breadcrumb) rather than a silent swallow, and still won't wedge the logout flow. Tests now exercise the open-handle + bare-handle delete paths through `wipeLocalDmStore`.

Issue #731, opened 2026-05-27
perf(messages): tab-switch jank — refreshDmInbox decrypt loop saturates the JS thread on Messages focus


BenGWeeks:
## Symptom

Janky/laggy when switching the bottom tab bar between **Messages**, **Explore**, and **Friends** on a physical Pixel (preview build). Persisted after #612 (Explore setState coalescing) and #595/#597 (live-location dot) — because those addressed different bottlenecks.

## Audit (Pixel, preview build, 7 Maestro-driven tab switches)

gfxinfo: p50 8ms / p90 13ms / p95 16ms / **p99 117ms**, 14/681 janky (2%). The pain is **all in the tail** — and **867 "high input latency" events (127% of frames)**: touch dispatch delayed on nearly every frame = the "laggy" feel. Three GC events (~50MB freed each) all coincide with Messages focus.

`[PerfTab]` tap→focus: Explore→Messages (cold) **245ms** worst; others 70–120ms.

## Root causes

1. **`refreshDmInbox` runs 7–11s on every Messages focus past the 30s TTL.** The NIP-17 decrypt loop saturates the JS thread. Captured: `refreshDmInbox: 10888ms, k1059=85, fresh=58`.
2. **`setTimeout(0)` yields inflate 5–20×under load.** `nostrDecryptPacing.ts` uses a 4ms budget; each wrap decrypt takes >4ms on this Pixel, so it yields *every wrap* (83 yields for 85 wraps), and each `setTimeout(0)` costs ~90ms under load (queue saturated by the loop re-enqueuing) → ~80% scheduler overhead.
3. **Aborted refreshes chain.** The 30s TTL (`dmInboxLastRefreshAt`) is written only when the refresh *resolves*; abort-on-blur leaves it unset, so the next Messages focus immediately starts another full refresh (tab-hop faster than the refresh → full refresh every time, indefinitely).
4. **GPS subscription restarts on every Explore/Friends focus+blur.** `useUserLocation` drives retain/release via `useFocusEffect`, but the screen is `freezeOnBlur`-frozen — so the watch tears down and restarts (`requestForegroundPermissionsAsync` fired 4× in the test). Regression introduced by #595/#597.
5. **(secondary) 27–46 Glide avatar re-decodes** (up to ~600ms each) on every Messages focus, decoded at intrinsic size (`-2147483648x-2147483648` = no size hint) → over-allocates bitmaps.

## Fixes (this issue)

[...]

PR #743, opened 2026-05-28
perf(messaging): skip-set for non-inbox NIP-17 wraps — warm refreshDmInbox ~16s → ~4s (#745)

### LINKED ISSUE #745: perf(messaging): warm refreshDmInbox 16 s — 28 non-inbox wraps re-decrypted every refresh
## Symptom

After **#741** landed, the tab-switch lockups are gone — but the **warm `refreshDmInbox`** is slow: **12–19 s wall-clock** on every Messages focus past the 30 s TTL, vs the pre-#733 baseline of **7–11 s**.

The refresh is now correctly **non-blocking** (#741 Fix 2 — RAF yields keep the JS thread responsive throughout, confirmed by `yields=57 (budget=4ms, cap=2)` in the perf log + UI stays interactive). So the user no longer *feels* it as a freeze. But the total time is elevated and the work is redundant.

## Root cause

From the post-#741 benchmark (Stev.ie, on the Pixel preview build):

> `[Perf] nip17-cache k1059=87 hits=59 misses=28`

87 kind-1059 wraps are fetched per warm refresh; 28 are "misses" — re-decrypted from scratch every single refresh.

Why those 28 don't get cached: the positive cache (`cache[wrap.id] = entry`) is only written when a decrypt *produces an inbox entry*. The 28 wraps decrypt successfully but `continue` before the cache write because they're:
- **Group rumors** — routed to group storage instead of the 1:1 inbox.
- **Messages from non-followed senders** — the B1 (follow-gate) `continue` fires before the cache write.

Both `continue` paths leave the wrap id with no cache record, so the next refresh re-fetches it and re-decrypts it identically, with no inbox-side change to show for the 4 ms × 28 wraps of NIP-44 work.

At ~4 ms per wrap × 28 wraps + 57 RAF yields × 4 ms = ~340 ms of pure waste per refresh (plus the relay round-trip for events the cache could have skipped).

## Proposed fix

A **negative-result skip-set**: a file-backed `Set<string>` of wrap ids where decrypt completed successfully but produced no inbox entry. Subsequent refreshes `Set.has(wrap.id)` short-circuit the decrypt entirely.

- File-backed (AsyncStorage), bounded (e.g. 1000 entries, LRU-trim like the existing positive cache).
- Both nsec and Amber decrypt branches must populate it.
- **Pull-to-refresh bypasses** the skip-set (with `!forceRefresh`) so a newly-followed contact's older wraps surface on the next manual refresh.
- Add `skipHits=N` to the `[Perf] nip17-cache` line so the saving is visible in logcat without a Perfetto trace.

## Expected impact

After the first warm refresh populates the skip-set, subsequent refreshes should drop **from ~16 s to ~4 s** (just the relay round-trip — no decrypt work for the 28 previously-skipped wraps).

## Out of scope

- Solving the `lastSeen` cursor lag (kind-1059 randomised `created_at` means the cursor doesn't always advance past the newest received event). Worth a separate look but orthogonal to this fix.

## Test plan

- Unit tests for the skip-set helpers (load empty, corrupt-JSON fallback, write + round-trip, cap enforcement, overwrite).
- Manual on the Pixel: open Messages cold (populates the set on first warm refresh), tap away + back several times, confirm subsequent warm refreshes log `skipHits=28` (or similar) and complete in ~4 s.
- Pull-to-refresh bypasses the skip-set (a previously-skipped wrap from a newly-followed contact resurfaces).

Builds on #731 / #733 / #741. Surfaced by Stev.ie's third benchmark run on the latest preview build.

BenGWeeks:
## What
Adds a **negative-result skip-set** for kind-1059 NIP-17 wraps that decrypt successfully but produce **no inbox entry** — i.e. wraps routed to group storage (group rumors) or dropped by the B1 follow-gate (non-followed senders). These ~28 wraps were being re-decrypted from scratch on every warm `refreshDmInbox`, contributing the bulk of the post-#741 12–19 s warm refresh time despite producing no observable inbox change.

A file-backed `Set<string>` of wrap ids is loaded at refresh start, `Set.has(wrap.id)` short-circuits the decrypt before NIP-44 work is done, and any newly-discovered no-entry decrypts are appended. **Pull-to-refresh bypasses** the skip-set (with `!forceRefresh`) so a newly-followed contact's older wraps still surface on the next manual refresh.

A `skipHits=N` counter is added to the `[Perf] nip17-cache` line so the saving is visible in logcat without a Perfetto trace.

Closes #745

## Files
- `src/contexts/nostrDmCache.ts` — skip-set constants + `loadNip17SkipSet` / `writeNip17SkipSet` (file-backed, bounded LRU-trim, mirrors the positive-cache pattern).
- `src/contexts/useDmInbox.ts` — skip-set load + check + write in both nsec and Amber branches; `forceRefresh` bypass; `nip17SkipHits` counter in the perf log.
- `src/contexts/nostrDmCacheSkipSet.test.ts` — 9 unit tests (empty file, corrupt JSON, round-trip, cap enforcement, overwrite).

## Expected impact
After the first warm refresh populates the set, subsequent warm refreshes drop from **~16 s → ~4 s** — just the relay round-trip remains, with no decrypt work for the 28 previously-skipped wraps.

Per the post-#741 benchmark (Stev.ie, on the Pixel preview build, issue #745): `[Perf] nip17-cache k1059=87 hits=59 misses=28` — those 28 are the targets.

## Test plan
- [x] `npx tsc --noEmit` clean.
- [x] ESLint clean, Prettier clean.
- [x] `bash scripts/check-file-size.sh` passes (no baselined file grew).
- [x] 9 new unit tests pass (`npx jest nostrDmCacheSkipSet`).
- [ ] Manual on the Pixel preview: open Messages cold (populates the set on the first warm refresh), tap away + back several times, confirm `skipHits=28` (or similar) in the `[Perf] nip17-cache` line and warm refresh wall-clock drops toward ~4 s.
- [ ] Pull-to-refresh — confirm a previously-skipped wrap from a newly-followed contact resurfaces (skip-set bypass works).

Builds on #731 / #733 / #741. Found by Stev.ie's third benchmark run on the latest preview build.


BenGWeeks:
Closing in favour of a clean rebase onto main (SendSheet size regression was from the co-merged pr423 commits, not this change). Reopening as a single commit on main.

BenGWeeks:
Superseded by #746 — cleaned up the pr423 bleed-in (the original branch accidentally contained two unrelated SendSheet commits because the Stevie agent worktree branched off pr423 instead of main). The skip-set work itself is unchanged; see #746 for the clean 3-file diff against current main.

Issue #846, opened 2026-06-11
perf(messages): cold-start DM decrypt sweep still freezes navigation 1–5s in the first minute (post-#789)


BenGWeeks:
## Symptom (measured, 2026-06-11)

Navigating the app in the first ~75s after a cold start stalls in 1–5s bursts — worst at Hunt sub-page enter/exit (3.5–5s), also Events enter and Messages tab focus. After the first circuit: zero >1s stalls across 19 further navigation circuits. On the Pixel's real inbox the sweep window is proportionally longer (52s previously measured), which presents as "navigating around the app starts freezing".

Full stress-test data (4 runs, 10–20 circuits, heartbeat-gap + gfxinfo instrumentation): see the measurement comment on #695 — https://github.com/BenGWeeks/lightning-piggy-mobile/issues/695#issuecomment-4684740748

## Why this is still open after #695 / #788 / #789

- The #695 SQLite store + #789 yield pacing shipped, but the once-per-session `refreshDmInbox` sweep still decrypts the backlog fresh on every cold start (measured: k4 fresh=212 + k1059 fresh=510 with warm caches preserved) and runs 28–30s on the JS thread.
- With `DECRYPT_YIELD_EVERY=15` (src/contexts/nostrDecryptPacing.ts), individual batches still block 1–1.5s when overlapped with Hunt/Events first-mounts — the yield granularity is too coarse for navigation overlap.
- PRs #825/#826/#831 measurably do NOT move these numbers (within noise) — different costs.

## Candidate directions

1. **Why are 212+510 events "fresh" on every cold start despite the persistent store?** Audit the hit path: if these should be cache hits, that's the bug; if they're genuinely new-since-cursor, the cursor window may be too wide.
2. **Cheap palliative:** drop the yield batch to ~5 during the full sweep (smaller blocks; total sweep lengthens but perceptible stalls shrink).
3. **Defer the sweep behind first interaction settle** (InteractionManager / idle callback) so circuit-1 navigation wins the thread.

Dev-build caveat: absolute gap sizes differ on release builds; gap locations should not.

PR #849, opened 2026-06-12
feat(messages): wire the encrypted DM store (SQLCipher) into the live inbox (#848)

### LINKED ISSUE #848: security(messages): wire the built-but-dormant encrypted DM store (SQLCipher) into the live inbox
## Problem

The #695/#690 encryption-at-rest machinery — `localDb.ts` (op-sqlite + SQLCipher), `localDbKey.ts` (256-bit key in the OS keystore, #697), `dmDb.ts` (indexed per-message rows), `dmIngest.ts`, `dmStoreMigration.ts` (migration from the file cache) — is fully built and unit-tested but **imported by nothing except its own tests** (found during the #846 investigation, see #847's root-cause analysis).

Consequence: decrypted DMs are persisted today as a plaintext JSON file (`nostrDmCache.ts`, `*_nip17_cache_v1_<pubkey>.json`, up to 12 MB) protected only by the OS sandbox / device FBE — Telegram-grade at-rest posture, while the Signal-grade store (SQLCipher + Keystore-held key, the model this project chose after reviewing how Signal does it) sits unplugged.

## Work

1. Wire `dmIngest`/`dmDb` into the live inbox path (`useDmInbox` / `nostrDmCache` consumers) so decrypted rumors persist to the encrypted store instead of the plaintext file.
2. Run `dmStoreMigration` once per account (file cache → DB, verified delete of the plaintext file afterwards — the migration module already implements this).
3. Keep the #743 skip-set semantics and the #847 backfill gating intact; reads stay indexed slices (never whole-inbox loads — that's the original #695 perf goal).
4. Logout/identity-switch wipe: ensure the #712 wipe primitives cover the DB rows.
5. Verify: second-cold-start fresh count stays ≈0 (the #847 gate must not regress); DMs survive the migration; the plaintext JSON file is gone from disk after migration.

Stacked on #847 (same files).

BenGWeeks:
## What this wires

The #695/#690 encryption-at-rest machinery — `localDb.ts` (op-sqlite + SQLCipher), `localDbKey.ts` (256-bit key in the OS keystore, #697), `dmDb.ts`, `dmIngest.ts`, `dmStoreMigration.ts` — was fully built and unit-tested but **imported by nothing except its own tests**. Decrypted DMs persisted as a plaintext JSON file (`*_nip17_cache_v1_<pubkey>.json`, observed at 228 KB / 510 entries on the test device). This PR plugs the store into every DM path and retires the plaintext cache:

- **`dmWrapIngest.ts` (new)** — the one NIP-17 inbox-wrap ingest engine, shared by both `refreshDmInbox` signer branches (nsec + Amber, kept symmetric) and the cold thread-open path. Gate order per wrap: encrypted-DB known-id → #743 skip-set → decrypt → group-route → B1 follow gate. Fresh rows batch-upsert into the DB; an aborted run persists neither rows nor skip-set growth. Amber `PERMISSION_NOT_GRANTED` still drains the remaining wraps without a dialog flood.
- **Kind-4 NIP-04 plaintexts now persist too** (inbox refresh, live sub, thread open) — the old NIP-04 cache was a RAM-only LRU that died with the session, so even a user pull-to-refresh that drops the `since` floor is now decrypt-once.
- **Owner-scoped schema** — rows are keyed `(owner, event_id)`, every query scoped by owner, so multi-account devices keep identities separate. The dormant pre-#848 table (dev installs only) is dropped and rebuilt.
- **Live sub** seeds its dedup Set from an id-only indexed query (`selectDmWrapIds`) instead of parsing the plaintext file; thread opens skip the inbox-wide relay fetch when the store is warm (`hasStoredWraps`).
- **Logout / identity-switch wipe** — `dmAccountWipe.ts` (file caches + skip-sets + that owner's DB rows + migration memo/flag) is called from `wipeAccountCaches`; the last-identity logout additionally deletes the DB file and its keystore key.
- `src/utils/nip17Cache.*` deleted; `useDmInbox.ts` shrinks 998 → 805 lines.

## Migration story

One-time, per-account, fired from every DM entry point (single-flighted + memoised). The old wrap-cache file was only ever a **memo of decryptions already performed**, so it is *imported* as rows — migration day costs one file read + one DB transaction, **not** a re-decrypt sweep (which is exactly what #846/#847 eliminated). Then the plaintext file is deleted with verification; strict ordering (populate → delete → verify → flag) is enforced by the pre-existing `dmStoreMigration` module. A failed run leaves plaintext + flag intact and retries on the next entry point. `docs/DATA_STORAGE.adoc` updated to match.

## Measurements (emulator dev client, 510-wrap / 2-conversation fixture inbox)

**Gate 1 — migration (cold start with plaintext cache present):**
- `[DmStore] migration: imported 510 wrap-cache entries into the encrypted DB (dropped 0 malformed)` → `migration complete … plaintext wrap cache deleted (verified)`
- `run-as` listing confirms `*_nip17_cache_v1_*.json` GONE from `files/`; `databases/lightningpiggy.db` (385 KB) created; DB header is ciphertext (not `SQLite format 3`) — SQLCipher active.
- First refresh after migration: `nip17-store: known=225, misses=0, stored=0`; thread open: `k1059 hits=5, fresh=0, skippedFetch=true`.

**Gate 2 — second cold start (the #847 gate):**
- No migration re-run (flag honored). Live sub seeds `510 dm + 37 group` wrap ids from the DB.
- Capped refresh: `known=225, misses=0, NIP-04 fresh=0`; backfill: `known=510, misses=0, stored=0` — **zero decrypts**, wall time is the async relay EOSE wait, same as #847.
- Messages tab paints both conversations from the DB; conversation opens correctly (Maestro, testIDs only).

**Gate 3 — circuit-1 stress (Home → Send open/close → Explore → Hunt in/out → Events in/out → Messages → Friends → Home), A/B on the *identical* flow + device state:**

| | base (#847 branch) | this PR (2 runs) |
|---|---|---|
| max heartbeat gap | 2,615 ms | 2,805 / 2,910 ms |
| Messages tab tap→focus | 1,805 ms | 1,991 ms |
| gaps >1 s | 8 | 4 / 6 |

Within emulator noise (~10%). The absolute numbers exceed #847's published 1.47 s because this circuit deep-scrolls Explore and enters/exits Hunt + Events (#847's circuit didn't); the max gap sits on the Explore→Messages tab transition with **zero decrypt or store work inside it** (store seed + live-sub open land after the gap, ~30 ms total). Incidentally, the base A/B run — whose file memo had been migrated away — paid a **48.6 s** refresh re-decrypting the whole inbox, demonstrating the fragility this PR removes: the decrypt-once memo now lives in an indexed, encrypted DB instead of a deletable JSON file.

## Test plan

- `npx jest` — full suite green (1,064 tests, 90 suites), incl. new co-located `dmWrapIngest.test.ts` + `dmStoreMigrationRunner.test.ts` and owner-scoping coverage in `dmDb` / `dmIngest` / `dmInbox` / `localDb` tests
- `npx tsc --noEmit`, ESLint (no new warnings), Prettier, `scripts/check-file-size.sh` — clean
- Emulator verification gates 1–3 above

**Stacked on #847 — merge that first.** (GitHub auto-retargets to `main` when #847 merges.)

Closes #848

## Review fixes (Archie's security review)

- **W1 — backup-restore brick**: wrong-key open failures (restored ciphertext + non-restored keystore key) now self-heal — bare-handle file delete + key clear + recreate; signature-gated so transients keep reject-and-retry and the SQLCipher-missing guard never heals (851f833, with heal/no-heal tests)
- **W2 — plaintext leak-back**: the inbox blob now persists `filteredFinal` (follow-gated), not `merged` (851f833)
- **N4 — wipe/migration race**: account wipe awaits an in-flight migration before deleting owner rows (851f833)
- **W3 + remaining nits** → follow-up #850: the plaintext conversation/inbox AsyncStorage blobs still duplicate DM plaintext outside SQLCipher — this PR closes the wrap-cache half of the at-rest story, not all of it.

## Scope (commits beyond the store wiring)

This PR grew during review/QA — reviewers should note it now carries three concerns, not one:

- `7050154` — wire the SQLCipher store into the live inbox (the original PR)
- `851f833` — Archie-review fixes (backup-restore self-heal, follow-gated persist, wipe/migration race)
- `39425d4` — **F1** (Teddie QA): post-login cold rebuild persists decrypted rows on abort + runs to completion, so the inbox isn't stranded empty
- `5fbc603` — **junk-conversation fix (#853)**: `partnerFromRumor` now HEX64-validates + lowercases the incoming sender pubkey (the `fromMe` branch already did); `buildDmSummaries` drops already-stored junk rows. **Security-adjacent** (sender identity — same area as #831's bind); +8 tests. *Folded in here rather than a standalone PR — flagged for an explicit review pass on retarget.*

All four together: 1,074 tests green; tsc / ESLint / Prettier / file-size clean.


BenGWeeks:
@copilot please re-review — added the F1 fix (39425d4): the post-logout cold DM rebuild now persists decrypted rows even on a tab-blur abort and runs to completion, with the dmIngest/dmWrapIngest abort-contract tests updated to match (rows persist on abort; inbox entries + skip-set stay suppressed). Worth a fresh look at the abort-signal handling in useDmInbox + dmIngest.

Issue #857, opened 2026-06-12
feat(messages): DM send outbox + retry — stop dropping messages that fail on first attempt


BenGWeeks:
## Problem (confirmed in code, 2026-06-12)

A DM that fails to publish on the first attempt is **dropped** — no retry, no queue, no persisted "failed" state:

- `nostrService.ts:1216` — each gift-wrap gets a single `await Promise.any(pool.publish(relays, wrap))`. One attempt; on flaky internet (all relays time out) it's recorded as an error and discarded. No backoff, no second try.
- `useConversationComposerActions.ts:58-67` — on `!result.success` the composer shows a one-time `Alert.alert('Send failed')` and `return false` **before** the optimistic local append, so the message is never added to the thread or cached. No "failed — tap to retry" bubble.
- **Partial-send trap:** each send is two wraps (recipient + self-copy). If the recipient wrap lands but the self-copy fails, the code returns failure → the recipient receives it but the sender's device drops it ("I sent it but don't see it"; the recipient has it fine). Reported by Ben on a poor connection.

## Fix direction (outbox pattern)

1. Persist the message locally *immediately* as a pending row (before/independent of publish).
2. Publish with retry + backoff; track per-relay acks (dovetails with the delivery-ticks issue).
3. On persistent failure, keep the bubble in a "failed — tap to retry" state rather than dropping it; offer manual + automatic retry (e.g. on reconnect / next app foreground).
4. Reconcile when the relay echo / self-copy finally lands.

Pairs with the delivery-ticks feature. Companion to that PR.

PR #865, opened 2026-06-12
fix(dm): white delivery ticks + timestamp on sent bubbles (#864)

### LINKED ISSUE #864: Sent-message delivery ticks and timestamp unreadable on pink bubble
On a sent (outgoing) DM bubble — which uses the brand-pink background — the delivery tick and the timestamp render in low-contrast colours and are hard to read:

- The **timestamp** shows in supplementary grey. The footer-row style `bubbleFooterTime` sets `color: textSupplementary`, which overrides the intended white `bubbleTimeMe` (a dead `bubbleFooterTimeMe` style was meant to fix this but was never wired up).
- The **delivery ticks** (single/double Check) use the payment-success green, which reads poorly on pink.
- The **pending** Clock uses supplementary grey — near-invisible on pink.

They should be white (as they used to be), since these elements only ever render on the pink sent bubble. The glyph alone carries the meaning (single vs double tick, clock, alert).

Failed sends keep the error red (distinct AlertCircle glyph) — intentional.

BenGWeeks:
## What

On a **sent** (outgoing) DM bubble — brand-pink background — the delivery tick and the timestamp were low-contrast and hard to read. This makes them white, the way they used to be.

- The footer-row `bubbleFooterTime` carried its own `color: textSupplementary` (grey), which overrode the intended white `bubbleTimeMe`. The style meant to supply the white override (`bubbleFooterTimeMe`) was defined but never wired into the component — dead code. Dropping the colour from `bubbleFooterTime` lets the composed `timeStyle` win: **white** on a sent bubble, supplementary grey on a received one.
- Delivered ticks (single/double `Check`) and the pending `Clock` now use white. They only ever render on the pink sent bubble, so the green/grey read poorly there. The glyph alone carries the meaning (single vs double tick, clock).
- **Failed** sends keep the error red (distinct `AlertCircle` glyph) — intentional, not a readability regression.

Styles-only change in `src/styles/MessageBubble.styles.ts`; no component logic touched.

## Why standalone

Folded out of #857 (optimistic-send flow, in progress) to avoid colliding with that branch's live work — this is a pure presentation fix and has no overlap with the send-flow changes.

## Test plan

- `npx tsc --noEmit` — clean
- `npx eslint` / `npx prettier --check` — clean
- `scripts/check-file-size.sh` — clean (file shrank)
- `npx jest` — 93 suites / 1107 tests green
- **On-device verification deferred** — visual check on the emulator (sent bubble: white time + white ticks legible on pink; received bubble: grey time unchanged) once the emulator frees up from #857.

Closes #864

### DIFF
--- a/src/styles/MessageBubble.styles.ts
+++ b/src/styles/MessageBubble.styles.ts
@@ -72,29 +72,25 @@ export const createMessageBubbleStyles = (colors: Palette) =>
       gap: 4,
     },
     // The time inside the footer row drops its own marginTop (the row owns it)
-    // and gains a small gap before the tick.
+    // and gains a small gap before the tick. It carries no colour of its own so
+    // the composed timeStyle wins — bubbleTimeMe (white) on a sent pink bubble,
+    // bubbleTime (supplementary grey) on a received surface bubble. (#864)
     bubbleFooterTime: {
       fontSize: 10,
-      color: colors.textSupplementary,
       marginRight: 4,
       // Zero the standalone bubbleTime top margin so the tick sits level with
       // the timestamp in the footer row (the row owns vertical spacing). (#858)
       marginTop: 0,
     },
-    bubbleFooterTimeMe: {
-      color: 'rgba(255,255,255,0.85)',
-    },
-    // Delivery tick (#856, design approved 2026-06-12). WhatsApp-style
-    // single/double coverage in the payment-success green (shared with
-    // PaymentProgressOverlay so the token matches across both themes):
-    // single Check = delivered to ≥1 relay, double CheckCheck = all relays.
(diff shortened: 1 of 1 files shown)


BenGWeeks:
## On-device before/after (added post-merge per request)

Captured on the Android emulator. Sent-bubble **timestamp goes grey → white**, now readable on the pink bubble:

Before/after: sent-bubble timestamp grey to white

Full after-state:

After: Big Piggy conversation with white timestamps

Note: the green ticks in the BEFORE crop don't appear in AFTER only because this fresh session carries no live delivery status on these historical messages (the persistent delivery store is #866, not yet on main). #865 recolours the ticks to white *when they render* — it doesn't remove them.

PR #872, opened 2026-06-13
fix(send): surface address-resolution error + edit address in place (#871)

### LINKED ISSUE #871: fix(send): unresolvable/ mistyped address is a silent dead-end — surface an error + allow editing the address in place
Reported on-device (emulator, dev build).

## Symptom
Entering a Lightning address with a typo (the address field also **drops characters while typing** — see Related) leaves the Send sheet showing `Pay to <mangled-address>` with **no amount control and a disabled, greyed-out Send button**, and **no error message**. It looks like a dead-end ("the amount disappeared and I can't try again"). The only recovery is the **"Scan / paste different invoice"** link, which **blanks the field** — so you cannot *edit* the existing address, you must retype it from scratch.

## Root cause (code)
`src/components/SendSheet.tsx`:
- A lud16 needs LNURL-pay params before an amount can be entered: `canSend = needsAmount ? (lnurlParams && currentSats > 0 && !resolving) : …` (~L790). When the address can't be resolved (typo / unreachable), `lnurlParams` stays `null` → `canSend` is `false` → Send disabled and the amount step never renders.
- **No error is surfaced** on a failed LNURL resolution — Send just silently disables. (Contrast: a genuine mid-send transport failure *is* handled and retry-able — `isConnectionError` → `connection-lost`, `handleOverlayDismiss` keeps the filled-in form, ~L678/L728.)
- The recovery affordance `handleReset()` (~L745) clears everything (`setPasteText('')`, `setSatsValue('')`, `setLnurlParams(null)`, …) → you re-enter blank; there's no "edit this address" path.

## Asks
1. **Surface a clear error** when an address/invoice can't be resolved (e.g. "Couldn't reach this Lightning address — check it and try again") instead of a silently-disabled Send.
2. **Allow editing the address in place** — tapping to fix the recipient should return to the input **pre-filled with the current value** (not blanked), so a one-character typo doesn't require retyping the whole address.
3. Keep the genuine post-send network-failure retry behaviour as-is (it already retains the form).

## Acceptance criteria
- [ ] An unresolvable Lightning address shows an inline/toast error, not just a disabled Send.
- [ ] The user can return to the address input with the previous value editable (fix-in-place), in addition to the existing "scan/paste different" reset.
- [ ] Verified on-device with a deliberately-mistyped Piggy lud16: error shown, address editable, correct send succeeds after fixing.

## Related
- Character-dropping on the Send address/paste field while typing (the trigger for the mistyped address) — likely the same input-corruption class seen on the amount/LNURL field. If not already tracked, worth a sibling issue.

BenGWeeks:
## What & why

An unresolvable / mistyped Lightning address used to be a **silent dead-end**: the Send sheet showed `Pay to <bad address>` with **no amount control and a disabled, greyed-out Send button — and no error message**. The only recovery, *"Scan / paste different invoice"*, **blanked the field**, so fixing a one-character typo meant retyping the whole address.

This PR makes resolution failure recoverable:

1. **Surface the error** — when `useSendSheetLnurl`'s resolution `.catch` fires, we now raise a friendly branded **`Toast`** ("Couldn't reach this Lightning address — check it for typos and try again") instead of leaving Send silently disabled. The raw error is logged, not shown.
2. **Edit in place** — the failure also bounces the user back to the editable input with the **bad value retained** (not blanked), so the typo can be fixed without retyping. A new **"Edit address"** affordance on the resolved details card offers the same fix-in-place recovery for an address that resolved but is still wrong. The existing full-reset *"Scan / paste different invoice"* link stays.
3. **No regression** to the genuine post-send transport-failure retry path (`connection-lost` / `error` → `handleOverlayDismiss` keeps the filled-in form) — untouched.

## File-by-file

- **`src/hooks/useSendSheetLnurl.ts`** — replaced the dead-end `Alert.alert('Error', …)` in both resolution catch blocks with a new `onResolveError(title, body)` callback (friendly copy; raw error `console.warn`'d). The lightning-address path previously left `scanned`/`invoiceData` set → the dead-end; it now hands control back to the sheet.
- **`src/components/SendSheet.tsx`** — added `handleEditAddress` (returns to the paste step, keeps `pasteText`/`activePubkey`) and `handleResolveError` (toast → edit); wired `onResolveError` into the hook; added an **"Edit address"** touchable plus `testID`/`accessibilityLabel` on the recovery links. Net **−25 lines** (1176 → 1151), so it stays **under its file-size baseline**.
- **`src/utils/sendSheetInput.ts`** — new pure helper `editAddressPrefill(pasteText, invoiceData)` (prefer the typed value, fall back to the parsed target, trim, never return `"null"`).
- **`src/utils/sendSheetInput.test.ts`** — 4 new cases for `editAddressPrefill`.
- **`tests/e2e/test-send-address-resolve-error-edit.yaml`** — pastes a deliberately-unresolvable lud16 (`piggy@nonexistent.invalid`, reserved RFC-2606 TLD), asserts the error toast shows, asserts the field is **pre-filled and editable** (append a char, confirm round-trip). testIDs only, no coordinates.

## Test plan

- [x] `npx eslint` (changed files) — clean (branded `Toast`, single-line comments)
- [x] `npx prettier --check` (changed files) — clean
- [x] `bash scripts/check-file-size.sh` — passes; **SendSheet.tsx 1151 ≤ 1176 baseline (did not grow)**
- [x] `npx jest` — **97 suites / 1152 tests pass**, including 4 new `editAddressPrefill` cases
- [ ] **On-device verification deferred to the maintainer** — the emulator is single-instance and currently busy; this branch was built in an isolated worktree without driving adb/Metro/emulator. Maestro flow `test-send-address-resolve-error-edit.yaml` is ready to run.

Closes #871

## Screenshots — on-device ✅

Captured on the Android emulator (this branch). A mistyped/unresolvable Lightning address used to be a silent dead-end (disabled Send, no error, recovery blanked the field). Now it surfaces a branded error and bounces back to the **editable field with the bad value retained**, so the typo is fixable in place.

**Before → After:**

Before: silent disabled-Send dead-end. After: error toast and editable pre-filled field.

**After — error toast + editable, pre-filled field:**

Couldn't reach this Lightning address toast over the editable, pre-filled address field

### DIFF
--- a/src/components/SendSheet.tsx
+++ b/src/components/SendSheet.tsx
@@ -11,6 +11,7 @@ import {
 } from 'react-native';
 import { Image as ExpoImage } from 'expo-image';
 import { Alert } from './BrandedAlert';
+import { Toast } from './BrandedToast';
 import {
   BottomSheetModal,
   BottomSheetBackdrop,
@@ -35,6 +36,7 @@ import { fetchInvoice, LnurlPayParams } from '../services/lnurlService';
 import {
   type DecodedInvoice,
   decodeInvoice,
+  editAddressPrefill,
   isLightningAddress,
   isValidInvoice,
   isLnurlString,
@@ -224,6 +226,35 @@ const SendSheet: React.FC<Props> = ({
     };
   }, []);
 
+  // Fix-in-place recovery (#871): return to the paste/input step with the bad
+  // value RETAINED (unlike handleReset, which blanks it) so a one-char typo
+  // can be corrected without retyping the whole address. Unwinds the
+  // resolved/scanned state but keeps pasteText / activePubkey.
+  const handleEditAddress = useCallback(() => {
+    const prefill = editAddressPrefill(pasteText, invoiceData);
+    setInvoiceData(null);
+    setDecoded(null);
+    setScanned(false);
+    setLnurlParams(null);
+    setResolving(false);
+    setSatsValue('');
+    setIsLnurl(false);
+    setIsOnchainAddress(false);
+    setStep('main');
+    setInputMode('paste');
+    setPasteText(prefill);
+  }, [pasteText, invoiceData]);
+
+  // Resolution failed (typo / unreachable): toast the friendly error, then
+  // hand the user straight back to the editable address (#871).
+  const handleResolveError = useCallback(
+    (title: string, body: string) => {
+      Toast.show({ type: 'error', text1: title, text2: body });
+      handleEditAddress();
+    },
+    [handleEditAddress],
+  );
+
   // Resolve a scanned/pasted lightning address or raw LNURL into LNURL-pay
   // params (or report a withdraw claim code). Extracted to keep SendSheet
   // under the file-size cap — see useSendSheetLnurl.
@@ -240,6 +271,7 @@ const SendSheet: React.FC<Props> = ({
     setScanned,
     setIsLnurl,
     setSatsValue,
+    onResolveError: handleResolveError,
   });
 
   const processInput = (data: string) => {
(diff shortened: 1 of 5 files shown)


BenGWeeks:
Addressed in bf87b3d. Both resolve effects now list `onResolveError` in their dependency arrays. To avoid the re-run loop this would otherwise cause (the callback chain closed over `pasteText`/`invoiceData`, so it changed on every keystroke), I made `onResolveError` reference-stable: `handleEditAddress` now reads the submitted value via render-synced refs instead of `useCallback` deps. A resolve failure still prefills the field with the value actually submitted (read live at failure time), never a stale one.

PR #876, opened 2026-06-13
fix(dm): wipe-path follow-ups — live-sub race, logout reset, stale drawer (#851)

### LINKED ISSUE #851: DM wipe-path follow-ups from #849 QA (live-sub race, rotted logout flow, stale drawer)
Minor findings from Teddie's QA of #849 (F1 is being fixed in #849 itself; these three are tracked here).

**F2 — Minor: live-sub follow-gate race after account switch.** Immediately after switching identity, a live wrap from the previous account is `dropped by follow-gate` (`src/contexts/nostrLiveDmSub.ts:412`) while follows are still hydrating. Recoverable (the drop isn't skip-set-persisted), but a genuinely fresh message arriving in that window loses its notification. Consider gating the live sub on follows-hydrated, or re-evaluating dropped wraps once hydration completes.

**F3 — Minor (pre-existing test debt): `tests/e2e/test-logout.yaml` is rotted.**
- `appId: ${APP_ID:-…}` crashes Maestro's JS evaluator ("Cannot read property 'lightningpiggy' of undefined")
- `tapOn: text: 'Sign Out'` is ambiguous against the branded alert (matches the title, leaves the dialog open) — should use `id: 'branded-alert-button-1'`
- the final assert still uses the removed `drawer-row-profile`

**F4 — Observation: drawer header showed "Big Piggy" after Big signed out** while logs showed the active session was already the fallback identity (`d9b33280`). Possibly stale drawer state on multi-account fallback; unrelated to storage.

Source: Teddie's QA pass on PR #849 (all 5 scenarios PASS; storage verified to SQL level).

BenGWeeks:
Addresses the three DM wipe-path follow-ups Teddie raised in #851 QA of #849.

### F2 — live-sub follow-gate race after account switch
Immediately after switching identity the contacts list hydrates async, so for a brief window `followPubkeysRef.current` is empty/stale and a genuinely-fresh inbound is dropped by the follow-gate (`nostrLiveDmSub.ts` kind-4 :220 / kind-1059 :413). The drop was recoverable via the next `refreshDmInbox` (never skip-set-persisted), but the **live inbox update + one-shot OS notification** for that arrival were lost until a manual refresh.

- New `src/contexts/liveSubFollowGate.ts`: a bounded (cap 50), per-sub deferral buffer holding the already-decrypted surface side-effects (inbox entry + optional notification) of fresh inbound dropped purely by the gate. No ciphertext, no persistence.
- The sub buffers fresh inbound it drops; `useDmInbox`'s `followPubkeys` effect calls `reevaluate` when follows hydrate, replaying surface + notify for partners that now pass.
- **Atomic with wipe:** sub teardown unregisters the replay hook (`setDeferredReplay(null)`) and clears the buffer, so a wipe / account switch can never replay a just-wiped wrap into the next identity. A late live callback after teardown still no-ops via the existing `cancelled` guard. Reuses the abort/bind-token teardown discipline already in the file.

### F4 — stale drawer header after sign-out
The logout-with-successor path promoted the successor (`setPubkey`/`setSignerType`/`setIsLoggedIn`) **without first clearing the previous identity's `profile`/`contacts`**, so the drawer header kept rendering the old display name ("Big Piggy") against the new active session — exactly the observed symptom.

- New `src/contexts/promoteSuccessorIdentity.ts`: teardown-before-set ordering (`setProfile(null)` + `setContacts([])` before promoting), plus a successor cached-profile hydrate so the header repaints to the successor immediately, plus a deferred relay refresh — mirroring `switchIdentity`.
- New `src/contexts/persistActiveIdentityKeys.ts`: the shared legacy single-identity SecureStore promotion, deduping `switchIdentity` and the logout path. Behaviour is byte-for-byte preserved (the `clearOtherSlot` flag captures the one difference between the two callers). Extracting these kept `NostrContext.tsx` **under** the file-size cap.

### F3 — rotted logout flow
The issue referenced `tests/e2e/test-logout.yaml`; that flow has since been reorganized to `.maestro/authentication/flow-007-logout.yaml`. Re-checking each of the issue's three F3 items against the **current** tree:
- ✅ **Fixed:** `tapOn: text: 'Sign Out'` was ambiguous against the branded alert title (matched the title, left the dialog open) → now `id: 'branded-alert-button-1'` (button order `[Cancel(0), Sign Out(1)]`).
- ☑️ **Already correct:** `appId: ${APP_ID || '…'}` is the standard expression across all 145 flows post-reorg (the rotted `${APP_ID:-…}` bash form was the old `tests/e2e/` copy).
- ☑️ **Already correct:** `drawer-row-profile` still exists (`AccountDrawerContent.tsx:49`) and is used by 15+ flows — it was removed in the old tree at QA time but has since been restored.

## Test plan
Unit tests (co-located in `src/contexts`, the Jest scope):
- `liveSubFollowGate.test.ts` — defer/reevaluate/clear, dedup by id, cap eviction, no double-replay, `clear()` drops everything (wipe).
- `nostrLiveDmSubFollowGate.test.ts` — teardown unregisters replay + clears buffer atomically; a late callback after teardown no-ops; a fresh non-followed wrap is buffered then replayed on hydration.
- `promoteSuccessorIdentity.test.ts` — clears stale profile/contacts before setting the successor (F4); hydrates successor caches; defers + converges the relay refresh.
- `persistActiveIdentityKeys.test.ts` — which SecureStore slots each signer writes + the `clearOtherSlot` difference.

Gates green locally: `tsc --noEmit` (clean), `eslint` (0 errors), `prettier --check` (clean), `scripts/check-file-size.sh` (pass — NostrContext back under cap), `jest` (106 suites / 1214 tests pass).

**On-device verification deferred** — a perf run is in progress on the main checkout, so adb/Metro/emulator were not used. The Maestro `flow-007-logout.yaml` fix and the runtime behaviour (live-sub replay window, drawer repaint on sign-out) should be validated on a device before release.

Does not regress the #849 encrypted-store wiring — the wipe/persist guards in `nostrLiveDmSub`/`dmAccountWipe` are untouched; the deferral buffer holds no plaintext and never persists.

Closes #851

### DIFF
--- a/.maestro/authentication/flow-007-logout.yaml
+++ b/.maestro/authentication/flow-007-logout.yaml
@@ -33,9 +33,12 @@ name: Nostr Logout Test
 - waitForAnimationToEnd:
     timeout: 2000
 
-# Confirm the Sign Out alert
+# Confirm the Sign Out alert. Tap the branded-alert destructive button by
+# id, NOT `text: 'Sign Out'` — the text matcher is ambiguous against the
+# alert's own title (also "Sign Out") and can match the title, leaving the
+# dialog open. Button order is [Cancel (0), Sign Out (1)], so index 1.
 - tapOn:
-    text: 'Sign Out'
+    id: 'branded-alert-button-1'
 - waitForAnimationToEnd:
     timeout: 3000
 
--- a/src/contexts/NostrContext.tsx
+++ b/src/contexts/NostrContext.tsx
@@ -1016,12 +1016,10 @@ export const NostrProvider: React.FC<{ children: React.ReactNode }> = ({ childre
         const { pubkey: pk } = nostrService.decodeNsec(trimmed);
         setPubkey(pk);
 
-        // Store credentials in the legacy single-active-identity slots
-        // (still the canonical location every other consumer reads from)
-        // AND register the identity in the multi-account store so the
-        // switcher knows it exists (#288).
-        await SecureStore.setItemAsync(NSEC_KEY, trimmed);
-        await SecureStore.setItemAsync(SIGNER_TYPE_KEY, 'nsec');
+        // Store credentials in the legacy single-active-identity slots via the
+        // canonical writer (hardened, device-only keychain) AND register the
+        // identity in the multi-account store so the switcher knows it (#288).
+        await persistActiveIdentityKeys({ pubkey: pk, signerType: 'nsec', nsec: trimmed });
         const next = await upsertIdentity({
           pubkey: pk,
           signerType: 'nsec',
@@ -1083,8 +1081,7 @@ export const NostrProvider: React.FC<{ children: React.ReactNode }> = ({ childre
       const pk = await amberService.requestPublicKey();
 
       setPubkey(pk);
-      await SecureStore.setItemAsync(PUBKEY_KEY, pk);
-      await SecureStore.setItemAsync(SIGNER_TYPE_KEY, 'amber');
+      await persistActiveIdentityKeys({ pubkey: pk, signerType: 'amber' });
       const next = await upsertIdentity({
         pubkey: pk,
         signerType: 'amber',
@@ -1252,6 +1249,7 @@ export const NostrProvider: React.FC<{ children: React.ReactNode }> = ({ childre
           setPubkey,
           setSignerType,
           setIsLoggedIn,
+          setDmInbox,
           loadProfileFromCache,
           loadContactsFromCache,
           hydrateDmInboxFromCache,
@@ -1287,6 +1285,7 @@ export const NostrProvider: React.FC<{ children: React.ReactNode }> = ({ childre
     loadRelays,
     loadProfile,
     hydrateDmInboxFromCache,
+    setDmInbox,
     setAmberNip44Permission,
     knownWrapIdsRef,
   ]);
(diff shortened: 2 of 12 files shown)


BenGWeeks:
Fixed in 7eedcf29. The shared `promoteSuccessorIdentity` helper now calls `setDmInbox([])` BEFORE promoting the successor (right after `setContacts([])`), so a successor with an empty cached inbox no longer inherits the prior identity's previews — `hydrateDmInboxFromCache`'s non-empty-only guard is now backstopped. `switchIdentity` already had its own `setDmInbox([])` teardown, so the gap was only in the logout-with-successor path; fixing the shared helper covers both call sites. New test in `promoteSuccessorIdentity.test.ts` asserts the empty-cache successor clears the inbox before set/hydrate.

BenGWeeks:
Hardened in 7eedcf29. Added `LEGACY_IDENTITY_SECURE_OPTIONS` (`keychainAccessible: AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY`) in `nostrAuthKeys.ts`, matching `identitiesStore.ts` / `walletStorageService.ts`. `persistActiveIdentityKeys` now passes it on every slot write, and the nsec + amber login flows in `NostrContext` (which previously wrote `NSEC_KEY`/`PUBKEY_KEY`/`SIGNER_TYPE_KEY` via bare `setItemAsync`) now route through the canonical `persistActiveIdentityKeys` writer instead — so nsec / identity metadata is no longer eligible for iCloud/backup migration on any write path. `persistActiveIdentityKeys.test.ts` asserts every slot is written with the hardened options.

BenGWeeks:
@copilot please re-review — addressed both threads in 7eedcf29.

Issue #598, opened 2026-05-17
fix(nav): cold-start always resets to Home — NavigationContainer doesn't persist state


BenGWeeks:
## Description

When the app is killed by the OS (more aggressively on GrapheneOS, but also stock Android under memory pressure, and iOS after a reboot or extended background) and the user reopens it, navigation **always** lands back on the Home tab. Whatever tab / screen / stack you were in is silently dropped.

User-visible symptom: *"I was on the Explore tab, I switched apps for five minutes, came back — I'm on Home again."*

## Root cause

`src/navigation/AppNavigator.tsx:348` mounts `NavigationContainer` with **no state persistence**:

```tsx
<NavigationContainer ref={navigationRef} theme={navTheme}>
```

Both `onStateChange` and `initialState` are missing. React Navigation's documented pattern for surviving a cold-start is:

1. On `onStateChange`, write the state to AsyncStorage.
2. On mount, read it back and pass as `initialState`.

Neither half is wired up. The bottom-tab navigator also has no `initialRouteName`, so it falls through to the first declared tab (Home) on every fresh render.

GrapheneOS killing background processes more aggressively only makes the bug more visible — the actual problem is that the app stores nothing about *where you were*.

## Expected behaviour

Reopening the app lands you on the **same tab / screen** you left it on, modulo:

- Deep-links (`lightningpiggy://...`, `lightning:...`, NFC taps) still take precedence — they should override the persisted state.
- Auth gate still wins — if the user signed out between sessions, the auth flow should still fire.
- Schema migration — if a referenced screen was renamed/removed in an app update, restoration should fail gracefully and fall back to defaults rather than crash.

## Suggested fix

Standard [React Navigation persistence pattern](https://reactnavigation.org/docs/state-persistence/):

1. Add `src/utils/navigationStatePersistence.ts` — small helper with `loadPersistedNavigationState()` / `persistNavigationState()` / `clearPersistedNavigationState()` backed by AsyncStorage and a versioned key.
2. In `AppNavigator.tsx`:
   - On mount, await `loadPersistedNavigationState()` once before rendering the container.
   - Pass the result as `initialState`.
   - Wire `onStateChange={(state) => persistNavigationState(state)}`.
   - Skip restore when `Linking.getInitialURL()` returns a value — let the existing deep-link handler in `App.tsx` win.
3. Catch JSON-parse / schema-mismatch errors → treat as "no saved state".
[...]

PR #887, opened 2026-06-13
fix(nav): stop replaying a stale action on cold start (#886)

### LINKED ISSUE #886: fix(nav): app replays a stale action on launch (Send sheet auto-opens to pay; cache page reopens)
## Symptom

On cold start the app re-enters a stale, one-shot action instead of starting clean:

- **Pixel (production):** the **Send sheet auto-opens pre-filled to pay a 500-sat invoice** (address shown as the invoice description, e.g. `element_comment:2648:publish`). Tapping Send would actually pay it — a real footgun / security concern.
- **iPhone (preview):** the app **reopens the last geo-cache (Piglet) page** ("the Allotment Watch") every launch.

## Root cause

Navigation-state persistence (#598) saves the **entire** nav state via `onStateChange`, including:
1. `HomeScreen`'s transient `sendToAddress` param (set by a Friends-tab zap; HomeScreen opens the Send sheet whenever it's present), and
2. transient deep-link destinations like `HuntPiggyDetail` / `HuntFound`.

On the next cold start `loadPersistedNavigationState()` restores that blob verbatim, so the one-shot action replays. The runtime param-cleanup in HomeScreen never reaches the already-persisted state.

## Fix

Sanitize the nav state in `src/utils/navigationStatePersistence.ts`, on **both save and load** (load is essential so devices already stuck self-heal on upgrade):
- strip one-shot action params (`sendToAddress` & friends) from every route, and
- pop transient deep-link routes (`HuntPiggyDetail` / `HuntFound`) so we restore the parent stack (ExploreHome / Hunt) instead of the leaf.

Affects all platforms. Not in v1.3.5.

BenGWeeks:
## What

On cold start the app re-entered a stale, one-shot action instead of starting clean:

- **Pixel (production):** the **Send sheet auto-opened pre-filled to pay a 500-sat invoice** — tapping Send would actually pay it (a real footgun / security concern).
- **iPhone (preview):** the app **reopened the last geo-cache (Piglet) page** every launch.

## Root cause

Navigation-state persistence (#598) saves the **entire** nav state via `onStateChange`, including:
1. `HomeScreen`'s transient `sendToAddress` param (set by a Friends-tab zap — HomeScreen opens the Send sheet whenever it's present), and
2. transient deep-link destinations (`HuntPiggyDetail` / `HuntFound`).

On the next cold start `loadPersistedNavigationState()` restores that blob verbatim, so the one-shot action replays. The runtime param-cleanup in HomeScreen never reaches the already-persisted state.

## Fix

`sanitizeNavigationState()` in `src/utils/navigationStatePersistence.ts`, applied on **both save and load**:
- strips one-shot action params (`sendToAddress`, `sendToPicture`, `sendToPubkey`, `sendToName`, `openComposer`) from every route, and
- pops transient deep-link routes (`HuntPiggyDetail` / `HuntFound`) — restoring the parent stack (ExploreHome / Hunt) instead of the leaf, clamping the index.

**Why sanitize on load too:** devices that already persisted a stale action (before this ships) self-heal on the next launch, instead of needing the fix *plus* a manual navigate-away. The function is pure, total, and never hands back an empty `routes` array (which would crash `NavigationContainer`).

Drive-by: fixed a pre-existing test-isolation bug in the same suite — the "swallows AsyncStorage failures" tests used `jest.spyOn(...).mockRestore()`, which drops the async-storage mock's in-memory backing for every later test. Switched to `mockRejectedValueOnce` on the mock fn (auto-reverts).

## Test plan

- `npx tsc --noEmit` — passes.
- `npx jest src/utils/navigationStatePersistence.test.ts` — 23 pass, incl. new cases: strips nested Home params, pops HuntPiggyDetail + clamps index, purity (no input mutation), never-empties-a-stack, and load()/persist() round-trips proving the heal-on-read and clean-on-write behaviour.
- Full suite: `npx jest` — 109 suites / 1250 tests pass.
- ESLint / Prettier — clean (0 errors).

Closes #886

### DIFF
--- a/src/utils/navigationStatePersistence.test.ts
+++ b/src/utils/navigationStatePersistence.test.ts
@@ -232,6 +232,13 @@ describe('sanitizeNavigationState — replay guard (#886)', () => {
     expect(sanitizeNavigationState({} as unknown as NavigationState)).toEqual({});
   });
 
+  it('collapses an empty-routes stack to undefined (no index -1 crash)', () => {
+    // A malformed-but-JSON-valid blob with zero routes would otherwise yield
+    // index = routes.length - 1 = -1 and crash NavigationContainer on mount.
+    const empty = { index: 0, routeNames: [], routes: [] } as unknown as NavigationState;
+    expect(sanitizeNavigationState(empty)).toBeUndefined();
+  });
+
   it('load() heals an already-persisted stale invoice (sanitize on read)', async () => {
     await AsyncStorage.setItem(
       NAV_STATE_KEY,
--- a/src/utils/navigationStatePersistence.test.ts
+++ b/src/utils/navigationStatePersistence.test.ts
@@ -5,6 +5,7 @@ import {
   clearPersistedNavigationState,
   loadPersistedNavigationState,
   persistNavigationState,
+  sanitizeNavigationState,
 } from './navigationStatePersistence';
 
 jest.mock('@react-native-async-storage/async-storage', () =>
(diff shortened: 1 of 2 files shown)


BenGWeeks:
Both addressed in 0c804649:
- **[blocking] empty-routes / index -1** — `sanitizeNavigationState` now collapses an empty `routes` array to `undefined` (navigator falls through to defaults) and clamps `index` into `[0, len-1]`. Added a test covering it.
- **[nit] purity doc** — reworded: the function is pure (no mutation) but returns a NEW cleaned object for valid states; only malformed/leaf states pass through by reference.

PR #919, opened 2026-06-20
fix(location): work without Google Play Services on GrapheneOS / AOSP (#918)

### LINKED ISSUE #918: compat(graphene): remove Google Play Services dependencies for GrapheneOS / de-Googled devices
## Goal

Make Lightning Piggy run cleanly on **GrapheneOS / de-Googled devices with no Google Play Services**. Several features currently depend on Google Play Services (GMS) code paths that fail (or silently no-op) when GMS is absent. This issue audits each dependency, the user-facing impact, and the recommended fix.

Audit performed against the current `main` (expo-location 55.1.8, expo-notifications 55.0.22).

---

## 1. Location — Play `SettingsClient` settings-check path (HIGHEST IMPACT, reproduced)

**Symptom:** On a device with no Play Services (GrapheneOS) and on a stock AOSP emulator, location features fail with:

> Location request failed due to unsatisfied device settings

**Root cause:** `expo-location`'s Android native module (`node_modules/expo-location/android/src/main/java/expo/modules/location/LocationModule.kt`) handles `getCurrentPositionAsync` like this:

- If the **`NETWORK_PROVIDER` is disabled** AND `mayShowUserSettingsDialog` is `true` (its **default**), it does **not** request a fix directly. It calls `resolveUserSettingsForRequest`, which uses `LocationServices.getSettingsClient(ctx).checkLocationSettings(...)` — the **Google Play `SettingsClient`** (`LocationModule.kt:474-499`).
- On a device with no Play Services that task fails with a status other than `RESOLUTION_REQUIRED`, so the module rejects with `LocationSettingsUnsatisfiedException` → the "unsatisfied device settings" string.
- A stock AOSP emulator typically has the network provider **off** (GPS-only, fed by `adb emu geo fix`), so this path is hit every time and location never resolves.

`watchPositionAsync` is **not** affected — it calls `requestLocationUpdates` directly and never touches the `SettingsClient`.

**Affected call sites (all one-shot `getCurrentPositionAsync`):**
- `src/services/locationService.ts` (`getCurrentLocation` → share-my-location DM)
- `src/services/geofenceService.ts` (`enableGeofencing` → nearby-merchant alerts)
- `src/hooks/useLiveUserLocation.ts` (live map dot, used by every map surface)
- `src/screens/PlaceDetailScreen.tsx`, `EventsScreen.tsx`, `MapScreen.tsx`, `PlacesScreen.tsx`, `ExploreHomeScreen.tsx`, `HuntCreateScreen.tsx`
- `src/components/LocationPickerSheet.tsx`

**User-facing impact on GrapheneOS:** Share-location, nearby-merchant geofencing, the live user dot, "pin here" when creating a Piglet, and the Places/Events/Map/Explore "centre on me" all fail with the error or never resolve.

**Recommended fix (being implemented in the linked draft PR):** Pass `mayShowUserSettingsDialog: false` to every one-shot `getCurrentPositionAsync`. That flag makes the native module skip the `SettingsClient` branch and request the fix directly via `requestSingleLocation`. `Accuracy.High` maps to `PRIORITY_HIGH_ACCURACY`, which prefers the raw GPS provider — exactly what `adb emu geo fix` feeds and what a de-Googled device exposes via the AOSP `LocationManager`. We lose the "turn on high-accuracy mode" system dialog, but that dialog is a Play surface that doesn't exist on GrapheneOS anyway. Centralised in a small unit-tested helper `src/services/aospLocation.ts`.

---

## 2. Push notifications — Firebase Cloud Messaging (FCM)

**Where:** `expo-notifications` ships an `ExpoFirebaseMessagingService` and a `FirebaseMessagingDelegate` (`node_modules/expo-notifications/android/src/main/java/expo/modules/notifications/service/...`). The Android config plugin is registered in `app.config.ts` (~line 137).

**Current status — GOOD:** The app deliberately uses **local notifications only**, not remote/FCM. `app.config.ts` documents this ("We rely on local (not remote) notifications only — no FCM token is requested", "OS notifications foundation … without FCM"), and `src/services/backgroundSyncService.ts` / `backgroundTask.ts` implement a detect-and-ping background sync via WorkManager/BGTaskScheduler rather than push.

**User-facing impact on GrapheneOS:** None for the intended local-notification feature. FCM would only matter if remote push were added later — which the project has explicitly rejected (see `docs/architecture/notifications.adoc`).

**Recommended action:** No code change needed. Optionally add a note/CI assertion that `getDevicePushTokenAsync` / `getExpoPushTokenAsync` (the FCM-token APIs) are never called, so a future change can't silently introduce an FCM dependency. Low priority.

---

## 3. Google ML Kit — barcode/QR scanning via expo-camera

**Where:** `app.config.ts` proguard keep-rules (~lines 210-213):
```
-keep class com.google.mlkit.** { *; }
-keep class com.google.android.gms.internal.mlkit_vision_** { *; }
-keep class com.google.android.gms.vision.** { *; }
```
These keep `expo-camera`'s MLKit barcode classes from being stripped by R8. `expo-camera` (`node_modules/expo-camera/android/build.gradle`) pulls the ML Kit barcode-scanning dependency, used for the in-app **QR scanner** (LNURL / invoice / npub scanning).

**Is it Play-coupled?** `expo-camera`'s `android/build.gradle` (lines 34-37) pulls **both**:
- `com.google.mlkit:barcode-scanning:17.3.0` — the **bundled** ML Kit model. Ships the detector + model in the APK and works **with no Play Services**.
- `com.google.android.gms:play-services-code-scanner:16.1.0` — Google's **Code Scanner UI**, which **is** Play-coupled (used only by expo-camera's `useGoogleCodeScanner` API).

The app's QR scanning (`src/components/SendScanPane.tsx`, `SendSheet.tsx`, `AddFriendSheet.tsx`) uses `CameraView` with `barcodeScannerSettings` + `onBarcodeScanned` — i.e. the **bundled ML Kit path**, NOT `useGoogleCodeScanner`. So QR scanning should work on GrapheneOS.

**User-facing impact on GrapheneOS:** Expected **none** for QR scanning, since the app uses the bundled detector. The Play `play-services-code-scanner` artifact is dead weight (unused API) but harmless at runtime. Worth confirming on-device that the bundled detector decodes without Play Services.

**Recommended action (follow-up):** (a) Confirm on a GrapheneOS device / no-GMS emulator that `onBarcodeScanned` decodes. (b) Optionally exclude the unused `play-services-code-scanner` dependency (set `expo.camera.barcode-scanner-enabled` handling or a Gradle exclusion) to drop a Play artifact from the APK. Low priority once decoding is confirmed working.

---

## 4. Maps — MapLibre (no Google Maps) — CONFIRMED FINE

`@maplibre/maplibre-react-native` is the map stack (`app.config.ts` plugins). There is **no `react-native-maps` / Google Maps dependency** in `package.json`. Vector tiles come from OpenFreeMap. No action needed.

---

## Summary table

| Item | Play-coupled? | GrapheneOS impact | Fix |
|------|---------------|-------------------|-----|
| 1. Location one-shot fix | **Yes** — `SettingsClient.checkLocationSettings` | **High, reproduced** — location fails | `mayShowUserSettingsDialog: false` (draft PR) |
| 2. Push / FCM | Latent only | None (local-only by design) | No change; optional guard |
| 3. ML Kit QR scan | Maybe (bundled vs unbundled) | QR scanner may not decode | Verify artifact; force bundled model |
| 4. Maps | No (MapLibre) | None | None |

The location fix (item 1) is the only one with reproduced breakage and ships as a draft PR linked to this issue. Items 2-3 are tracked here as follow-ups.

BenGWeeks:
## What & why

On **GrapheneOS / de-Googled devices** and on the **bare AOSP emulator**, one-shot location requests failed with:

> Location request failed due to unsatisfied device settings

That string is expo-location's `LocationSettingsUnsatisfiedException`. Walking the native module (`expo-location/android/.../LocationModule.kt`): when the `NETWORK_PROVIDER` is disabled **and** `mayShowUserSettingsDialog` is `true` (its default), `getCurrentPositionAsync` does **not** request a fix directly — it calls `resolveUserSettingsForRequest`, which uses the **Google Play `SettingsClient.checkLocationSettings()`**. On a device with no Play Services that task fails, so the module rejects. A stock AOSP emulator is GPS-only (network provider off, fed by `adb emu geo fix`), so this path is hit every time and location never resolves.

`watchPositionAsync` was never affected — it calls `requestLocationUpdates` directly and never touches the `SettingsClient`. Only one-shot `getCurrentPositionAsync` callers broke.

## The fix

Pass `mayShowUserSettingsDialog: false` to every one-shot fix. With that flag the native module skips the `SettingsClient` branch and requests the fix directly via `requestSingleLocation`. `Accuracy.High` maps to `PRIORITY_HIGH_ACCURACY`, which prefers the raw GPS provider — exactly what `adb emu geo fix` feeds and what a de-Googled device exposes via the AOSP `LocationManager`. The lost "turn on high-accuracy" system dialog is a Play surface that doesn't exist on GrapheneOS anyway.

Centralised in a small, unit-tested helper **`src/services/aospLocation.ts`** (`getOneShotPosition` / `oneShotPositionOptions` / `DEFAULT_LOCATION_ACCURACY`) and threaded through every one-shot caller:

- `src/services/locationService.ts`, `src/services/geofenceService.ts`
- `src/hooks/useLiveUserLocation.ts`
- `src/screens/PlaceDetailScreen.tsx`, `EventsScreen.tsx`, `MapScreen.tsx`, `PlacesScreen.tsx`, `ExploreHomeScreen.tsx`, `HuntCreateScreen.tsx`
- `src/components/LocationPickerSheet.tsx`

The over-cap `HuntCreateScreen.tsx` was folded by two lines so the new import leaves it net-smaller (file-size gate green).

This is **item 1** (the only reproduced breakage) of the GrapheneOS audit in #918. Items 2 (FCM — already local-only, no change) and 3 (ML Kit QR — app uses the bundled detector, likely fine; exclude the unused Play `code-scanner` artifact) remain as follow-ups in that issue.

## Test plan

- [x] `npx eslint` on all changed files — 0 errors (pre-existing warnings only)
- [x] `npx prettier --check` on changed files — clean
- [x] `bash scripts/check-file-size.sh` — passed (no baselined file grew)
- [x] `npx jest` — 116 suites / 1298 tests pass, incl. new `src/services/aospLocation.test.ts` (pins `mayShowUserSettingsDialog: false`)
- [ ] **On-device / emulator (no Play Services):** start a bare AOSP emulator, stream `adb emu geo fix <lon> <lat>`, open a map surface (Explore / Places / Map) → the user dot should now resolve and follow the injected fix instead of failing with "unsatisfied device settings". Also verify share-location and "pin here" (Hunt create).

Closes #918

### DIFF
--- a/src/components/LocationPickerSheet.tsx
+++ b/src/components/LocationPickerSheet.tsx
@@ -9,6 +9,7 @@ import {
   ActivityIndicator,
 } from 'react-native';
 import * as Location from 'expo-location';
+import { getOneShotPosition } from '../services/aospLocation';
 import {
   BottomSheetModal,
   BottomSheetBackdrop,
@@ -174,7 +175,8 @@ const LocationPickerSheet: React.FC<Props> = ({
       .then((perm) => {
         if (cancelled) return;
         if (perm.status !== 'granted') return; // last-known/UK paths take over
-        return Location.getCurrentPositionAsync({ accuracy: Location.Accuracy.High });
+        // No-Play-Services-safe one-shot fix (see aospLocation.ts).
+        return getOneShotPosition();
       })
       .then((pos) => {
         if (!pos || cancelled) return;
--- a/src/hooks/useLiveUserLocation.ts
+++ b/src/hooks/useLiveUserLocation.ts
@@ -1,5 +1,6 @@
 import { useEffect, useRef, useState } from 'react';
 import * as Location from 'expo-location';
+import { oneShotPositionOptions } from '../services/aospLocation';
 
 /**
  * Live user-location subscription for map views.
@@ -155,7 +156,11 @@ export function useLiveUserLocation(
       // already bit us in `locationService.ts`) silently block the
       // live subscription forever — so the dot would freeze on the
       // last-known fix and never update.
-      Location.getCurrentPositionAsync({ accuracy })
+      // `oneShotPositionOptions` adds `mayShowUserSettingsDialog: false` so
+      // the one-shot fix never hits the Play `SettingsClient` path that
+      // rejects with "unsatisfied device settings" on de-Googled devices.
+      // See src/services/aospLocation.ts.
+      Location.getCurrentPositionAsync(oneShotPositionOptions(accuracy))
         .then((fresh) => {
           if (cancelled) return;
           applyIfNewer(fresh);
--- a/src/screens/EventsScreen.tsx
+++ b/src/screens/EventsScreen.tsx
@@ -13,6 +13,7 @@ import {
   RefreshControl,
 } from 'react-native';
 import * as Location from 'expo-location';
+import { getOneShotPosition } from '../services/aospLocation';
 import {
   CalendarDays,
   ChevronLeft,
@@ -179,9 +180,8 @@ const EventsScreen: React.FC<Props> = ({ navigation }) => {
           setLoading(false);
           return;
         }
-        const liveFix = await Location.getCurrentPositionAsync({
-          accuracy: Location.Accuracy.High,
-        });
+        // No-Play-Services-safe one-shot fix (see aospLocation.ts).
+        const liveFix = await getOneShotPosition();
         // Focus was lost while we awaited location — bail before opening a
         // sub the cleanup already ran past.
         if (cancelled()) return;
--- a/src/screens/ExploreHomeScreen.tsx
+++ b/src/screens/ExploreHomeScreen.tsx
@@ -9,6 +9,7 @@ import {
   RefreshControl,
 } from 'react-native';
 import * as Location from 'expo-location';
+import { getOneShotPosition } from '../services/aospLocation';
 import { useFocusEffect } from '@react-navigation/native';
 import {
   CalendarDays,
@@ -172,9 +173,8 @@ const ExploreHomeScreen: React.FC<Props> = ({ navigation }) => {
         // Non-fatal — fall through to getCurrentPositionAsync.
       }
       try {
-        const current = await Location.getCurrentPositionAsync({
-          accuracy: Location.Accuracy.High,
-        });
+        // No-Play-Services-safe one-shot fix (see aospLocation.ts).
+        const current = await getOneShotPosition();
         if (!cancelled) {
           setPos({
             lat: current.coords.latitude,
--- a/src/screens/HuntCreateScreen.tsx
+++ b/src/screens/HuntCreateScreen.tsx
@@ -16,6 +16,7 @@ import { CameraView, useCameraPermissions } from 'expo-camera';
 import * as Clipboard from 'expo-clipboard';
 import * as ImagePicker from 'expo-image-picker';
 import * as Location from 'expo-location';
+import { getOneShotPosition } from '../services/aospLocation';
 // SAF + base64 helpers live on the legacy expo-file-system module. The
 // new class API (File/Paths) doesn't expose StorageAccessFramework yet,
 // so we mix the two: class API for the cache copy, legacy for the SAF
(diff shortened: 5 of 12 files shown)


BenGWeeks:
Closing — this was a misdiagnosis. The "unsatisfied device settings" error came from the **emulator** (a Google Play image) having Location Accuracy turned off + the Play accuracy prompt being declined, not from a GrapheneOS/app bug. On GrapheneOS (no Play Services) expo-location falls back to the AOSP LocationManager and works fine — confirmed on a real GrapheneOS Pixel. The proposed `mayShowUserSettingsDialog: false` change would also suppress the helpful 'turn on location accuracy' prompt on stock Google devices, so it's net-negative. Correct fix for the emulator is to enable location accuracy there, no code change needed.

PR #1093, opened 2026-09-16
perf(home): isolate payment sheet rendering (#1092)

### LINKED ISSUE #1092: Reduce cold-start and Receive/Send render work on Home
The app can feel slow immediately after launch, and Receive/Send taps can take seconds to respond.

Home currently mounts all five sheet components during startup. Their visibility state also lives in HomeScreen, so opening a sheet rerenders the wallet carousel and transaction list before the sheet can present.

Move sheet state into a separate host and defer each sheet's initial mount until requested. Keep sheets mounted after first use so dismiss effects and in-flight payments survive closing. Verify parent/sibling render isolation and retention with regression tests; add timing markers for device follow-up. This addresses avoidable render work, not all possible JS-thread startup stalls.

BenGWeeks:
Receive/Send visibility lived in HomeScreen, so each tap rerendered the wallet carousel and transaction list before presenting the requested sheet. Home also mounted all five sheet components during cold start, even when none had been opened.

Move sheet state into a memoized HomeSheets host. Mount each sheet on first use, then keep it mounted across closes so dismissal effects and in-flight payment/recovery operations survive. Memoized sibling sheets receive stable callbacks, and navigation-provided send targets still open through the same host.

Render-work comparison (verified in the host regression harness with native sheets mocked):

| Operation | Before | After |
| --- | --- | --- |
| Sheet trees mounted on Home cold start | 5 | 0 |
| Home parent updates caused by a Receive/Send open | 1 | 0 |
| Previously mounted Receive rerenders when opening Send | 1 | 0 |

The host adds `[PerfBlock] render:HomeSheets` timings, and both payment buttons log their tap dispatch. This removes avoidable render work; it does not establish a device-level millisecond improvement or prove that all reported multi-second stalls have been resolved. Module imports remain eager. No Android bridge/device was available locally.

## Test plan

- Five host regressions pass: cold mounting, parent/sibling isolation, retained state, all five flows, immediate navigation target, and stable ref commands.
- Full Jest suite, TypeScript, repository ESLint (existing warnings), changed-file Prettier, file-size/performance gates, and diff checks passed.
- Device follow-up: cold-launch with cached wallets; immediately tap Receive, close, and tap Send. Compare existing HomeScreen `[PerfBlock]` timings and new HomeSheets timings on base/head. Home must not commit solely because of either tap. Repeat with slow relays, a camera-enabled Send flow, a Friends zap target, Transfer, Add Wallet, and Settings. Confirm closing an in-flight payment does not cancel its background completion.

Closes #1092


BenGWeeks:
@coderabbitai review

BenGWeeks:
@coderabbitai review

